PR #1, opened 2017-09-25
ParseConfig doesn't understand -Wl flags.

merlin66:
This issue was originally created at: 2004-05-03 01:46:30.
This issue was reported by: `noselasd`.
noselasd said at 2004-05-03 01:46:30
>This has annoyed me abit. ParseConfig(..) is nice to parse output of
>xxx-config programs and pkg-config. However ParseConfig doesn't understand
>the -Wl flag. Which means 'pass flag to linker'. It ends up in the CPPDEFINES
>as far as I can see.
>Thus on a machine here, every pkg-config --libs will have among others
>'-Wl,-R/usr/pkg/lib' . And as ParseConfig places that in the CPPDEFINES, it
>should go to LINKFLAGS I guess. I get a warning on compiling every .c file that
>linking isn't done, and linking might fail since it actually doesn't get all
>link flags.
>I guess this goes for other -Wl flags as well, afaik ther's -Wsomething to pass 
>flags to the assembler and preprocessor as well.

stevenknight said at 2004-05-03 07:01:11
>Fix has been checked in for next release.  Thanks!

stevenknight said at 2006-01-05 18:14:20

kmaples said at 2006-05-20 17:03:30

### DIFF
--- a/test/Deprecated/Options/chdir.py
+++ b/test/Deprecated/Options/chdir.py
@@ -43,6 +43,7 @@ SConscript('subdir/SConscript')
 """)
 
 SConscript_contents = """\
+from __future__ import print_function
 Import("opts")
 env = Environment()
 opts.Update(env)
--- a/test/Deprecated/Options/import.py
+++ b/test/Deprecated/Options/import.py
@@ -45,6 +45,7 @@ SConscript('subdir/SConscript')
 """)
 
 SConscript_contents = """\
+from __future__ import print_function
 Import("opts")
 env = Environment()
 opts.Update(env)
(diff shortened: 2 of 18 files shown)


bdbaddog:
I think the tests can be fixed without the future. The issue is this:
```
print("VARIABLE =", repr(env['VARIABLE']))
#   the comma above makes it a tuple that's being printed under py 2.7
```

merlin66:
Would this variant using + be ok?
`print("VARIABLE =" + repr(env['VARIABLE']))`

I've got a bunch of edits to redo this, mainly +, but in some cases (where value is not string, I switched to %s"%  I'm going to decline this and push mine.

bdbaddog:
See: https://github.com/SConsProject/scons/pull/2

merlin66:
Ok, thanks, then I will let you handle it, just let me know if there's something I should fix.

PR #9, opened 2017-10-06
Search Mailing List Archive Does not work.

ztessler:
This issue was originally created at: 2006-01-24 01:33:19.
This issue was reported by: `shlomif`.
shlomif said at 2006-01-24 01:33:19
>The search mailing list archive: 
> 
>http://xrl.us/jper 
> 
>Does not work regardless of what query one inputs there. 
> 
>Please fix it. 
> 
>Regards, 
> 
>       Shlomi Fish

kmaples said at 2006-05-20 17:03:30

stevenknight said at 2006-12-08 15:55:12
>Not sure what mailing list archive that is, but it doesn't look like ours.

More information about this issue is at http://scons.tigris.org/servlets/SearchList?list=users&searchText=find&defaultField=body&Search=Search.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -26,6 +26,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       This was introduced by a pull request to allow recursive variable evaluations to yield a string
       such as "$( $( some stuff $) $)".
 
+  From Zachary Tessler:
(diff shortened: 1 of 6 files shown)


bdbaddog:
Can you add a test where the override is not the same?

ztessler:
Added two new tests, one for the case where the environments are different, and one for where the overrides are different. These test for behavior that isn't changed by this PR, and both pass on master as well.

PR #10, opened 2017-10-07
Clean up Node.FS classes.

dmoody256:
This issue was originally created at: 2001-07-18 22:00:00.
stevenknight said at 2001-07-18 22:00:00
>Requirements sent via email.

issues@scons said at 2001-07-18 22:00:00
>Converted from SourceForge task item 34780

stevenknight said at 2006-05-20 20:47:53
>No white space in keyword.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -11,6 +11,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
     - Whatever John Doe did.
 
+  From Daniel Moody:
+    - Updated the Jar Builder tool in Tool/__init.py so that is doesn't force class files as
+      sources, allowing directories to be passed, which was causing test/Java/JAR.py to fail.
+
   From William Deegan:
     - Fix issue where code in utility routine to_String_for_subst() had code whose result was never
       properly returned.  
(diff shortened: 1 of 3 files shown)


dmoody256:
No failed Java tests with Mint 17 JDK 1.8 for python 2.7 and python 3.5

bdbaddog:
Does it work with your changes if the sources are class files?

dmoody256:
The JAR.py test has some other tests which pass class files directly to the Jar builder, also I tested that independently with one of my projects.

bdbaddog:
Here's the manpage for the Jar builder:

```
Jar() , env.Jar()
Builds a Java archive (.jar) file from the specified list of sources. Any directories in the source list will be searched for .class files). Any .java files in the source list will be compiled to .class files by calling the Java Builder.

If the $JARCHDIR value is set, the jar command will change to the specified directory using the -C option. If $JARCHDIR is not set explicitly, SCons will use the top of any subdirectory tree in which Java .class were built by the Java Builder.

If the contents any of the source files begin with the string Manifest-Version, the file is assumed to be a manifest and is passed to the jar command with the m option set.

env.Jar(target = 'foo.jar', source = 'classes')

env.Jar(target = 'bar.jar',
        source = ['bar1.java', 'bar2.java'])
```

So the jar builder should accept both directories (which should get searched for .class files), or java files which should get compiled to class files.

dmoody256:
Regexing the Java tests with `grep -rPzo 'Jar.*.\[Jj][Aa][Vv][Aa]'` and `grep -rPzo '.\[Jj][Aa][Vv][Aa].*Jar'` only turns up one test (multi-step.py) which tests for Jar taking in .java as a source. 

Maybe a source with .java should be in the JAR.py test for completeness. Other than that, I think this fixes the JAR builder.

dmoody256:
Added another piece to the JAR.py test to test for java source files as a source to the Jar builder. Ran the tests again with JDK 1.8, python2.7 and python3.5.

PR #11, opened 2017-10-10
User guide

rzr:
This issue was originally created at: 2001-07-18 22:00:00.
stevenknight said at 2001-07-18 22:00:00
>Get user documentation to the point where someone will be able to use the initial release to compile simple C programs.

issues@scons said at 2001-07-18 22:00:00
>Converted from SourceForge task item 34779

### DIFF
--- a/src/engine/SCons/Taskmaster.py
+++ b/src/engine/SCons/Taskmaster.py
@@ -1016,6 +1016,27 @@ class Taskmaster(object):
                 to_visit = to_visit | parents
                 pending_children = pending_children - parents
 
+                # Clean up the other branches of side effect
+                for se in node.get_side_effects():
+                    waiting_s_e = se.get_waiting_s_e()
+                    pending_children = pending_children - waiting_s_e
+                    stack = set(waiting_s_e)
+                    if T:
+                        for n in waiting_s_e:
+                            T.write(self.trace_message('       removing side effect parent %s from the pending children set\n' %
+                                    self.trace_node(n)))
+                    while stack:
+                        n = stack.pop()
+                        waiting_parents = n.waiting_parents
+                        n.waiting_parents = set()
+                        stack = stack | waiting_parents
+                        pending_children = pending_children - waiting_parents
+
+                    for p in waiting_parents:
+                        p.ref_count = p.ref_count - 1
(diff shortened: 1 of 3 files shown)


bdbaddog:
Please add a blurb to src/CHANGES.txt

src/engine/SCons/Taskmaster.py
```diff
@@ -1016,6 +1016,27 @@ def will_not_build(self, nodes, node_func=lambda n: None):
                 to_visit = to_visit | parents
                 pending_children = pending_children - parents
 
+                # Clean up the other branches of side effect
+                for se in node.get_side_effects():
+                    waiting_s_e = se.get_waiting_s_e()
+                    pending_children = pending_children - waiting_s_e
+                    stack = set(waiting_s_e)
+                    if T:
+                        for n in waiting_s_e:
```
garyo, line 1025: Minor: could use `stack` instead of `waiting_s_e` here to avoid printing duplicates in the trace.

src/engine/SCons/Taskmaster.py
```diff
@@ -1016,6 +1016,27 @@ def will_not_build(self, nodes, node_func=lambda n: None):
...
+                            T.write(self.trace_message('       removing side effect parent %s from the pending children set\n' %
+                                    self.trace_node(n)))
+                    while stack:
+                        n = stack.pop()
+                        waiting_parents = n.waiting_parents
+                        n.waiting_parents = set()
+                        stack = stack | waiting_parents
```
garyo, line 1032: I thought about whether a dependency cycle could cause an infinite loop here, but I think it's safe -- because if `n` gets back on the list, its `waiting_parents` will already be null.

src/engine/SCons/Taskmaster.py
```diff
@@ -1016,6 +1016,27 @@ def will_not_build(self, nodes, node_func=lambda n: None):
...
+                        pending_children = pending_children - waiting_parents
+
+                    for p in waiting_parents:
```
garyo, line 1035: So here, `waiting_parents` is the list from the last node in the stack above. I'm not sure, but maybe the above `while stack` loop should be accumulating _all_ the waiting parents from all nodes in the stack?

PR #17, opened 2017-10-26
Add Depends() method.

dmoody256:
>Tracking explicit dependencies will be easiest first.

>Converted from SourceForge task item 34792

stevenknight said at 2006-05-20 20:49:15
>No white space in keyword.

### DIFF
--- a/src/engine/SCons/Tool/__init__.py
+++ b/src/engine/SCons/Tool/__init__.py
@@ -913,15 +913,25 @@ def createCFileBuilders(env):
 #  Create common Java builders
 
 def CreateJarBuilder(env):
+    """The Jar builder expects a list of class files
+    which it can package into a jar file.
+
+    The jar tool provides an interface for passing other types
+    of java files such as .java, directories or swig interfaces
+    and will build them to class files in which it can package
+    into the jar.
+    """
     try:
-        java_jar = env['BUILDERS']['Jar']
+        java_jar = env['BUILDERS']['JarFile']
     except KeyError:
         fs = SCons.Node.FS.get_default_fs()
         jar_com = SCons.Action.Action('$JARCOM', '$JARCOMSTR')
         java_jar = SCons.Builder.Builder(action = jar_com,
                                          suffix = '$JARSUFFIX',
+                                         src_suffix = '$JAVACLASSSUFFIX',
+                                         src_builder = 'JavaClassFile',
                                          source_factory = fs.Entry)
-        env['BUILDERS']['Jar'] = java_jar
+        env['BUILDERS']['JarFile'] = java_jar
     return java_jar
 
 def CreateJavaHBuilder(env):
--- a/src/engine/SCons/Tool/jar.py
+++ b/src/engine/SCons/Tool/jar.py
@@ -95,10 +95,22 @@ def Jar(env, target = None, source = [], *args, **kw):
     Builders.
     """
 
+    # mutiple targets pass so build each target the same from the 
+    # same source
+    #TODO Maybe this should only be done once, and the result copied
+    #     for each target since it should result in the same?
+    if SCons.Util.is_List(target) and SCons.Util.is_List(source):
+        jars = []
+        for single_target in target:
+            jars += env.Jar( target = single_target, source = source, *args, **kw)
+        return jars
+
     # jar target should not be a list so assume they passed
     # no target and want implicit target to be made and the arg
     # was actaully the list of sources
-    if SCons.Util.is_List(target):
+    if SCons.Util.is_List(target) and source == []:
+        SCons.Warning.Warning("Making implicit target jar file, " +
+                              "and treating the list as sources")
         source = target
         target = None
 
@@ -117,10 +129,6 @@ def Jar(env, target = None, source = [], *args, **kw):
         target = [target]
     if not SCons.Util.is_List(source):
         source = [source]
-
-    # Pad the target list with repetitions of the last element in the
-    # list so we have a target for every source element.
-    target = target + ([target[-1]] * (len(source) - len(target)))
 
     # setup for checking through all the sources and handle accordingly
     java_class_suffix = env.subst('$JAVACLASSSUFFIX')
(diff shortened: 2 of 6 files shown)


src/engine/SCons/Tool/jar.py
```diff
@@ -87,10 +89,115 @@ def jarFlags(target, source, env, for_signature):
...
+def Jar(env, target = None, source = [], *args, **kw):
+    """
+    A pseudo-Builder wrapper around the separate Jar sources{File,Dir}
+    Builders.
+    """
+
+    # jar target should not be a list so assume they passed
+    # no target and want implicit target to be made and the arg
+    # was actaully the list of sources
+    if SCons.Util.is_List(target):
```
bdbaddog: In this case I'm thinking a warning should be issued rather than quietly discarding the target specified by the user?

src/engine/SCons/Tool/__init__.py
```diff
@@ -913,15 +913,25 @@ def createCFileBuilders(env):
...
+    which it can package into a jar file.
+
+    The jar tool provides an interface for passing other types
+    of java files such as .java, directories or swig interfaces
+    and will build them to class files in which it can package
+    into the jar.
+    """
     try:
-        java_jar = env['BUILDERS']['Jar']
+        java_jar = env['BUILDERS']['JarFile']
```
bdbaddog, line 925: Did you change the name of the builder here from Jar to JarFile?

src/engine/SCons/Tool/jar.py
dmoody256: I agree, warnings for when things are getting weird is good, I'll update with a commit on this request soon.

Also on this note, is there any expected behavior when you pass multiple targets to the Jar builder? I didn't see anything in the man page. In this case, I assumed this behavior from the Java multi-step.py test, where it passes only a list of sources to the Jar builder: https://github.com/SConsProject/scons/blob/a583f043aec89e58bf4ab0ab20cc039299eff1df/test/Java/multi-step.py#L377

src/engine/SCons/Tool/__init__.py
dmoody256: Yes, I added a builder wrapper in its place with the AddMethod function: 
https://github.com/SConsProject/scons/pull/17/commits/b54f4166d0053ae7168af1d16223fff574344f17#diff-06522ed2191b2caac07035c96d8cbfa4R199
I took this design from the Java builder which has a similar design pattern for taking in sources from both directories and java sources, see below for comparison:
https://github.com/SConsProject/scons/blob/a583f043aec89e58bf4ab0ab20cc039299eff1df/src/engine/SCons/Tool/__init__.py#L954
https://github.com/SConsProject/scons/blob/a583f043aec89e58bf4ab0ab20cc039299eff1df/src/engine/SCons/Tool/javac.py#L163-L167

src/engine/SCons/Tool/jar.py
bdbaddog: Hopefully expected behavior is covered in the docs and/or the test suite.

Probably worth putting some notice that you'll be dropping the target list in the release notes (src/CHANGES.txt)?

Do any of the tests hit this case?

bdbaddog: I'm trying to roll out 3.0.1 and would like to get this change in. Any chance you can do the above today?

dmoody256: >Hopefully expected behavior is covered in the docs and/or the test suite.

The man page for Jar says it will build "a" jar archive so I assume that implies a single target:
http://scons.org/doc/HTML/scons-man.html
```
Jar() , env.Jar()
Builds a Java archive (.jar) file from the specified list of sources.
```
>Probably worth putting some notice that you'll be dropping the target list in the release notes (src/CHANGES.txt)?

I will update the CHANGES.txt about the target list not being used.

>Do any of the tests hit this case?

No tests test multiple targets passed to Jar, I can write a test for that but I still am not sure what is expected in that case. Currently it assumes if your first argument is a list (base builder class first arg is target), then you meant implicit target and just passed the list of sources, so if someone was trying to pass a list of targets and a list of sources it would ignore the sources and assumes your targets were the sources. Maybe this case should be an Error and stop the build?

>Any chance you can do the above today?

I can update regarding the warning message and the CHANGES.txt in a few hours.

dmoody256:
Updated and build passing here:
https://travis-ci.org/dmoody256/scons/builds/301730660

src/engine/SCons/Tool/jar.py
```diff
@@ -98,18 +98,30 @@ def Jar(env, target = None, source = [], *args, **kw):
     # jar target should not be a list so assume they passed
     # no target and want implicit target to be made and the arg
     # was actaully the list of sources
-    if SCons.Util.is_List(target):
+    if SCons.Util.is_List(target) and source == []:
```
bdbaddog: In general all values returned from builders are NodeLists. So it's possible a user could specify a target as a list, but that list could have only one item in it. Perhaps better to also check if the list has more than one item here?

bdbaddog: Never mind. Your solution seems reasonable given a target list with one or more targets.

bdbaddog:
Looks like this broke someone's usage. 
Can you take a look at:
https://pairlist4.pair.net/pipermail/scons-users/2017-November/006469.html

And respond on users list?

dmoody256:
@bdbaddog

Yeah, it's because they passed a list of lists when they made a list from the manifest and the list of class files. If they extend the class file list with the manifest instead it should work. However, I think the Jar builder should be able to handle this case.


I propose the Jar builder breaks down the lists of lists from the source arg into a single list and uses the single list in place? Thoughts?





_Sent from my Galaxy Note5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

bdbaddog:
Looks like this broke someone's usage. 
Can you take a look at:
https://pairlist4.pair.net/pipermail/scons-users/2017-November/006469.html

And respond on users list?

PR #18, opened 2017-11-18
Build Program from intermediate .o file nodes

dmoody256:
>Take the nodes returned from one or more Object() calls and feed them to Program() to create the final executable.

>Converted from SourceForge task item 34795

stevenknight said at 2006-05-20 20:49:26
>No white space in keyword.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 dist: trusty
 before_install:
-  - sudo apt-get -y install clang gdc docbook-xml xsltproc libxml2-dev libxslt-dev python-pip python-dev fop docbook-xsl-doc-pdf texlive-full biber texmaker build-essential libpcre3-dev autoconf automake libtool bison subversion git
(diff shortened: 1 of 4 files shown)


dmoody256:
Travis build is here:
https://travis-ci.org/dmoody256/scons/builds/303888167

mwichmann (commented):
Haven't checked if the flattening works since am now flattening in our own code, but it looks reasonable.  The manifest problem is not related; I believe the problem there is that an os.path.isfile() when a variant dir is in effect will fail the check for the filename of the manifest. The test looks like it should catch when it's gone wrong (that is, jar command has added a default manifest because it didn't think one was specified)

dmoody256:
Closing this pull request and resubmitting with scons flatten instead of re-implementing

dmoody256:
resubmitted with scons flatten here: https://github.com/SConsProject/scons/pull/19

PR #22, opened 2017-11-23
Scanner for C files

dmoody256:
This issue was originally created at: 2001-07-30 22:00:00.
stevenknight said at 2001-07-30 22:00:00
>Create a Scanner object that looks for #include lines in C/C++ files.

issues@scons said at 2001-07-30 22:00:00
>Converted from SourceForge task item 35419

stevenknight said at 2006-05-20 20:49:51
>No white space in keyword.

### DIFF
--- a/src/engine/SCons/JobTests.py
+++ b/src/engine/SCons/JobTests.py
@@ -27,17 +27,40 @@ import random
 import math
 import sys
 import time
+import os
 
 import TestUnit
 
 import SCons.Job
 
 
+def get_cpu_nums():
+    # Linux, Unix and MacOS:
+    if hasattr( os, "sysconf" ):
+        if "SC_NPROCESSORS_ONLN" in os.sysconf_names:
+            # Linux & Unix:
+            ncpus = os.sysconf( "SC_NPROCESSORS_ONLN" )
+        if isinstance(ncpus, int) and ncpus > 0:
+            return ncpus
+        else: # OSX:
+            return int( os.popen2( "sysctl -n hw.ncpu")[1].read() )
+    # Windows:
+    if "NUMBER_OF_PROCESSORS" in os.environ:
+        ncpus = int( os.environ[ "NUMBER_OF_PROCESSORS" ] );
+    if ncpus > 0:
+        return ncpus
+    return 1 # Default
+
 # a large number
 num_sines = 10000
 
 # how many parallel jobs to perform for the test
-num_jobs = 11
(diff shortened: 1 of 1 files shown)


dmoody256:
For reference:
https://pairlist2.pair.net/pipermail/scons-dev/2017-November/004556.html

> I was investigating why I was getting intermittent failures for JobTest.py
> on some Travis CI builds.
> 
> On my local 8 core machine, I found that playing around with the num_jobs
> in the JobTest.py would cause the test to fail. It was roughly (although
> not exactly?) related to if the num_jobs I chose was less than the number
> of cores available.
> 
> the current value is hard coded to 11 (maybe back when 16 or 32 cores was
> not technically available?). Maybe this should detect the number of cores
> and double it?
> 
> Looking at the code it seems that it could be possible that it finishes the
> job faster then it takes to spin up a new thread?
> 
> I saw this behavior with both python 2.7 and 3.5. Can any one else test on
> some different number of core machines and see if they get similar results?
> Or does anyone have ideas what is happening here?

bdbaddog:
test failed under travis.. can you take a look? 
https://travis-ci.org/SCons/scons/builds/312236981?utm_source=email&utm_medium=notification

dmoody256:
I tried to reproduce this, but I could not:
https://github.com/dmoody256/scons/commits/JobTestsFailing

It seems to intermittently fail, I was able to repro it locally on my machine by changing the job number, and it seemed that increase the job number would ensure better results. I am not sure why its failing on travis in some cases.

If you have the paid version of Travis you can setup debugging on the system to investigate issues failing in travis.

dmoody256:
Also it looks like you can troubleshoot travis tests in a docker image as well:
https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image

When I have some more time I will get a local image and see if I can repro and test this

bdbaddog:
Thanks!

dmoody256:
the build you linked is passed now, did you rerun it?

dmoody256:
A recent build not related to SCons TaskMaster or Jobs I did had another intermittent JobTests fail:
https://travis-ci.org/dmoody256/scons/builds/313214836

and then basically the same commit passed in another build:
https://travis-ci.org/dmoody256/scons/builds/313219015

bdbaddog:
Yes. I reran.
I'm guessing it's intermittent and passes/fails depending on the host/vm it
lands on and/or the loading of the underlying hardware?

PR #24, opened 2017-12-03
Push-button installation

ajf58:
This issue was originally created at: 2001-07-30 22:00:00.
stevenknight said at 2001-07-30 22:00:00
>Be able to install and use SCons with minimal hassle, whether from .rpm, .deb,  or straight from .tar.gz.

issues@scons said at 2001-07-30 22:00:00
>Converted from SourceForge task item 35382

stevenknight said at 2006-05-20 20:50:23
>No white space in keyword.


bdbaddog:
Are there any tests which are now also obsolete?

PR #27, opened 2017-12-08
Create Debian package installation.

dmoody256:
This issue was originally created at: 2001-08-28 22:00:00.
stevenknight said at 2001-08-28 22:00:00
>Make sure we can create .deb files as well.  Ideally, this would be done through distutils, but that didn't  support .deb natively the last time I looked.

issues@scons said at 2001-08-28 22:00:00
>Converted from SourceForge task item 37153

stevenknight said at 2006-05-20 20:51:02
>No white space in keyword.


dmoody256:
Adding this workaround to another pull request which also includes coverage testing

PR #28, opened 2017-12-09
Add -d (dependency print) support.

dmoody256:
This issue was originally created at: 2001-09-12 22:00:00.
stevenknight said at 2001-09-12 22:00:00
>After we have dependencies coded.

issues@scons said at 2001-09-12 22:00:00
>Converted from SourceForge task item 37985

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ before_install:
   - tar xzf rel-3.0.12.tar.gz
   - cd swig-rel-3.0.12 && ./autogen.sh && ./configure --prefix=/usr && make && sudo make install && cd ..
   - sudo pip install coverage
-  - sudo echo "import coverage" >> /usr/lib/python2.7/sitecustomize.py
-  - sudo echo "coverage.process_startup()" >> /usr/lib/python2.7/sitecustomize.py 
+  - echo "import coverage" | sudo tee --append /usr/lib/python2.7/sitecustomize.py
+  - echo "coverage.process_startup()" | sudo tee --append /usr/lib/python2.7/sitecustomize.py 
 
 script: 
   - export COVERAGE_PROCESS_START=$PWD/.coveragerc
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,11 @@ script:
   - echo "[run]" >> .coveragerc
   - echo "source = $PWD/src" >> .coveragerc
   - echo "parallel = True" >> .coveragerc
-  - coverage run --rcfile=$PWD/.coveragerc runtest.py -a || if [[ $? == 2 ]]; then true; else false; fi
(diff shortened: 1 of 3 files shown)


bdbaddog:
Any way to have travis agregate the test results so we can see them all in one place rather than having to check all the sub jobs?

dmoody256:
Not with the coverage tests, the coverage basically doubled the execution time. One option is to do a plain runtest.py -a without coverage in another job alongside the coverage jobs.

Data is also completely separate from each job unless you have some external source: https://docs.travis-ci.com/user/build-stages/#Data-persistence-between-stages-and-jobs

bdbaddog:
Do the jobs run in the same sandbox? (Share filesystem?)

dmoody256:
They run in docker images and can't access anything outside of their container.

I'm updating the Travis script to run the test suite without coverage in one job so all the tests are there.

.travis.yml
```diff
@@ -24,35 +12,69 @@ before_install:
...
+      before_script: skip
+      after_success: skip
+
+    - &coverage_jobs
+      stage: Coverage
+      
+      before_script:
+        - sudo pip install coverage
+        - sudo pip install coveralls
+        - echo "import coverage" | sudo tee --append /usr/lib/python2.7/sitecustomize.py
```
bdbaddog, line 33: Can we updated runtest.py to import coverage if available? Would that be better?

dmoody256: We would need to make some modifications to the SconsTest modules so they start their subprocesses with coverage enabled. This Travis script does it the quick and easy way but does have the drawback of needing to modify the python distribution.

For the time being I think this is good for getting coverage reports going for the project, while the only benefit to making the SconsTest modules use coverage directly would be the convenience of not needing to mess with the python distribution.

This is how coverage actually recommends you monitor subprocesses as well:
https://coverage.readthedocs.io/en/coverage-4.4.2/subprocess.html#
> Complex test suites may spawn sub-processes to run tests, either to run them in parallel, or because sub-process behavior is an important part of the system under test. Measuring coverage in those sub-processes can be tricky because you have to modify the code spawning the process to invoke coverage.py.
> 
> There’s an easier way to do it: coverage.py includes a function, coverage.process_startup() designed to be invoked when Python starts. It examines the COVERAGE_PROCESS_START environment variable, and if it is set, begins coverage measurement. The environment variable’s value will be used as the name of the configuration file to use.

bdbaddog: I'd be interested in being able to run coverage via runtest.py without having to run via travis/coveralls... But not needed to merge this logic. Thanks for the work!

bdbaddog:
Anyway to change from Ruby to Python in the travis scons page?

PR #34, opened 2017-12-23
Changes to address failing tests in appveyor

dmoody256:
Original text lost from flawed tigris bug migration

### DIFF
--- a/test/Clang/clang_default_environment.py
+++ b/test/Clang/clang_default_environment.py
@@ -25,7 +25,7 @@
 __revision__ = "__FILE__ __REVISION__ __DATE__ __DEVELOPER__"
 
 import TestSCons
-from SCons.Environment import Base
+import sys
 
 _exe = TestSCons._exe
 test = TestSCons.TestSCons()
@@ -33,8 +33,7 @@ test = TestSCons.TestSCons()
 if not test.where_is('clang'):
     test.skip_test("Could not find 'clang', skipping test.\n")
 
-platform = Base()['PLATFORM']
-if 'win32' == platform:
+if 'win32' == sys.platform:
     test.skip_test("clang is not part of default environment on windows, skipping test.\n")
 
 ##  This will likely NOT use clang
--- a/test/Clang/clang_default_environment.py
+++ b/test/Clang/clang_default_environment.py
@@ -25,12 +25,17 @@
 __revision__ = "__FILE__ __REVISION__ __DATE__ __DEVELOPER__"
 
 import TestSCons
+from SCons.Environment import Base
 
 _exe = TestSCons._exe
 test = TestSCons.TestSCons()
 
 if not test.where_is('clang'):
     test.skip_test("Could not find 'clang', skipping test.\n")
+
+platform = Base()['PLATFORM']
+if 'win32' == platform:
+    test.skip_test("clang is not part of default environment on windows, skipping test.\n")
 
 ##  This will likely NOT use clang
 
--- a/test/Clang/clang_shared_library.py
+++ b/test/Clang/clang_shared_library.py
@@ -33,6 +33,8 @@ test = TestSCons.TestSCons()
 if not test.where_is('clang'):
     test.skip_test("Could not find 'clang', skipping test.\n")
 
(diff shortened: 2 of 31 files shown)


test/Clang/clang_default_environment.py
```diff
@@ -25,13 +25,18 @@
...
 import TestSCons
+from SCons.Environment import Base
 
 _exe = TestSCons._exe
 test = TestSCons.TestSCons()
 
 if not test.where_is('clang'):
     test.skip_test("Could not find 'clang', skipping test.\n")
 
+platform = Base()['PLATFORM']
```
bdbaddog: Doesn't the above where_is('clang') take care of this?

test/Clang/clang_shared_library.py
```diff
@@ -41,15 +43,18 @@
     filename = 'foo.os'
     libraryname = 'libfoo.dylib'
 elif platform == 'win32':
-    filename = 'foo.obj'
+    filename = 'foo.os'
     libraryname = 'foo.dll'
+    # add the environment, otherwise the environment will consist of only vcvarsall.bat variables
+    # and not clang
+    env_str = "import os\nenv = Environment(tools=['clang', 'link'], ENV = os.environ)"
```
bdbaddog, line 49: Why do this only for win32?

test/Clang/clang_specific_environment.py
```diff
@@ -25,17 +25,25 @@
...
+from SCons.Environment import Base
 
 _exe = TestSCons._exe
 test = TestSCons.TestSCons()
 
 if not test.where_is('clang'):
     test.skip_test("Could not find 'clang', skipping test.\n")
 
+env_str = "env = Environment(tools=['clang', 'link'])"
+platform = Base()['PLATFORM']
```
bdbaddog: better to use sys.platform == for this rather than adding additional dependency on Base.. which can change or be removed.

test/Clang/clang_default_environment.py
dmoody256: default environment on windows will have the vcvarsall.bat run which will not taken into account the original environment and path that test.where_is sees

bdbaddog:
Your changes will disable and/or break tests which do complete successfully on our windows buildbot workers and on properly configured machines.

See: http://buildbot.scons.org/#/builders/8/builds/19

Plus I really don't like huge pull requests which affect many different tests.  Ideally you'd address the comments and make sure that the changes aren't disabling tests which work outside the appveyor environment.

The goal is not that only the tests which function in the appveyor environment run, but rather that tests which can pass will run in any environment which can run them and get skipped in those that cannot.

test/Clang/clang_shared_library.py
dmoody256: The Environment on windows was created without the original environment that the runtest.py was called from. Not sure if that is an appveyor specific issue or something weird with the msvs tool running vcvarsall.bat, but the mingw tools for linking were missing from the path causing a build failure on the link step during the test.

Adding the os.environ here retained the original environment and path that runtest.py was run from.

test/Clang/clang_specific_environment.py
dmoody256: agreed, I used the scons way for consistency with other tests, but I would prefer sys.platform.

dmoody256:
> Your changes will disable and/or break tests which do complete successfully on our windows buildbot workers and on properly configured machines.
>
>See: http://buildbot.scons.org/#/builders/8/builds/19

None of the failed test in the buildbot you linked were modified in this PR, and I dont see any underlying files in this PR which could have affected those tests on buildbot.

> Plus I really don't like huge pull requests which affect many different tests. Ideally you'd address the comments and make sure that the changes aren't disabling tests which work outside the appveyor environment.
>
> The goal is not that only the tests which function in the appveyor environment run, but rather that tests which can pass will run in any environment which can run them and get skipped in those that cannot.

I addressed your review comments, if there are certain changes in this PR you feel are changing the tests to disable or break them in other environments, we could check one of the appveyor environment variables and skip the test in those cases for appveyor so as to leave the test the same on other environments, but still be able to get passing results from appveyor.

PR #36, opened 2017-12-30
Add -W support

dmoody256:
This issue was originally created at: 2001-09-23 22:00:00.
stevenknight said at 2001-09-23 22:00:00
>The extremely cool GNU what-if option, showing what would be built if a file was changed.  We should be able to do this easily by just adding a random (or hard-coded) value to the signature.

issues@scons said at 2001-09-23 22:00:00
>Converted from SourceForge task item 39031

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,13 +53,12 @@ jobs:
         - sudo pip install coveralls
         # set this ensure user sites are available
         - export PYTHONNOUSERSITE=
-        # we need to preserve the test directories because they have the coverage data
-        - export PRESERVE=1
-        # attempt to get a location where we can store the usercustomize.py fiel
+        # attempt to get a location where we can store the usercustomize.py file
         - python -m site
         - export PYSITEDIR=$(python -m site --user-site)
         - sudo mkdir -p $PYSITEDIR
         - sudo touch ${PYSITEDIR}/usercustomize.py
+        - export COVERAGE_FILE=$PWD/.coverage_file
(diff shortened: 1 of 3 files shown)


bdbaddog:
wouldn't setting COVERAGE_FILE be simpler?

http://coverage.readthedocs.io/en/coverage-4.2/cmd.html#data-file

> Coverage.py collects execution data in a file called ”.coverage”. If need be, you can set a new file name with the COVERAGE_FILE environment variable. This can include a path to another directory.

dmoody256:
It seemed that wouldn't work with the multi-process run, because that section also says:
> By default, each run of your program starts with an empty data set.

And the next section talks about combining coverage data from multiple processes:
https://coverage.readthedocs.io/en/coverage-4.2/cmd.html#combining-data-files
> If you need to collect coverage data from different machines or processes, coverage.py can combine multiple files into one for reporting.

I didn't try using the COVERAGE_FILE file because it didn't seem that multiple processes writing to one file was how it was intended to work due to concurrency issues of which files have io access or which data gets written when.

bdbaddog:
hmm... o.k .I think it might work if you don't run runtest.py -j (>1)...

Trying to simplify the underlying changes and also not make changes when
not running via coverage or travis,etc..
It should still  work as it did when running on your own machine.
(though supporting coverage on a local test is good, but it shouldn't be
mandated)

dmoody256:
I tested the coverage_file env var and it worked, I'll update the travis script to use that as it is a bit simpler

PR #3045, opened 2018-01-11
 update mtime on cache retrieval

gabrielrussell:
With this patch when regular files are retrieved from the cache, the cached files mtime is updated. This is to allow one keep a cache on a disk that's mounted noatime, and then to still be able to prune old files.

If it's not possible to update the mtime on the cached file, because the filesystem is read-only, for example, then it simply does nothing.

One caveat is that the mtime is not updated for symlinks. This shouldn't be much of a problem being that symlinks don't consume much disk space and additionally should be easy to recreate.


bdbaddog:
Can you add a blurb to src/CHANGES.txt
Do you have a script which uses mtime updates to prune the cachedir which you could contribute?

bdbaddog:
The test is covering the permission failure mode. Can you test the success mode? (verify the mtime is updated on specified cached file)?

PR #3047, opened 2018-01-16
Call vulcan from scons and removed graph environment option

ppitta-spirent:
1) Graph is not set to default as "scons.gv"
2) Introduced new command line option "--mode" with options as local and vulcan. If the vulcan option is set then the scons will trigger the vulcan after generating the graph.
3) New environment variables are added for mode, vulcan options and path to vulcan exe


bdbaddog:
What's this?

ppitta-spirent:
Please ignore this PR. This was wrongly sent to the SCons:master

PR #3050, opened 2018-01-17
Coveralls badge not pointing to master branch

dmoody256:
Updated the README so the Coveralls badge is pointing to the master branch, otherwise it points to the most recent build.

also made a small change to the alt text so it will say build like the travis alt text.


bdbaddog:
Good Catch!

PR #3051, opened 2018-01-17
Do not add unnormalised entries to PATH in scons.bat

mingwandroid:


bdbaddog:
So what happens with out this? Python crashes?
All pythons, or only some versions?
Can you add a blurb to src/CHANGES.txt indicating the above.

mingwandroid:
I added a comment to the commit referencing the bug I filed at bpo: 
https://bugs.python.org/issue32457
Basically Python can not locate itself and therefore any of its stdlib and exits swiftly.

I'll try to get round to it soon but I'm pretty busy.

bdbaddog:
Looks like from the python bug it only affects PY 3.6, 3.7 and 3.8?

mingwandroid:
The problem here is to do with expectations of what `PATH` will contain. At no point does Windows normalize `PATH` so you put the onus on various executables (since Scons is a build system) to be able to handle this for themselves. This is a risky proposition and it's a needless variable to introduce. My PR addresses that cleanly.

bdbaddog:
The logic in scons.bat has been there a very long time, so clearly something has changed in Python which is now causing an issue with the generated paths.

Can you respond to my questions and add a blurb to src/CHANGES.txt?
( The request for this is in the pull request template which you should have seen when you made the pull request? )

We can't merge it without understanding which python versions this is an issue with.
Also we'd like to attribute changes to those who create them by the entries in src/CHANGES.txt

mingwandroid:
I tested it on Python 3.6 only. Does your CI not exercise `scons.bat`? As I said, I'm very busy, I'll try to make an entry to src/CHANGES.txt sometime, but I didn't already because I didn't think this change warranted it. FWIW, I've already applied a patch locally to fix this to the SCons that we distribute.

mingwandroid:
And whether something changed in python or not is not really important. I repeat, Windows never normalises PATH, so putting `..` in there is something that the executables that get called need to deal with. Since you cannot know what those executables are it is best not to assume they can handle it.

mingwandroid:
OK done. I removed the comment from scons.bat and put it in src/CHANGES.txt

bdbaddog:
Thanks!

Issue #3042, opened 2018-01-02
java manifest file is missed if passed as a string


bdbaddog:
This issue was originally created at: 2017-12-16 09:41:34.
This issue was reported by: `mwichmann`.
mwichmann said at 2017-12-16 09:41:34
>At the moment (3.0.1, github master which contains further changes to jar.py),
>if the invocation of env.Jar() contains a manifest file in the source list, it
>is fine if passed as a node but is ignored if passed as a string.
> 
>I'm not completely sure how jarManifest gets called, but it looks like, by the
>time it is called, the list of sources does not include the manifest due to this
>sequence:
> 
>in the Jar function, there is a loop through the original source list. If an
>entry in that list is already a node, it is processed and added to target_nodes
>(which is why it works if the manifest is passed as a node).  If it is not we
>check if it's a class file, a class directory, a java file. If not those, we
>check if it's a swig file. If not that, then:
> 
>                else:
>                    # The final else case handles anything not caught above and
>makes
>                    # sure other nodes that are sources for this jar get add as
>                    # a source to the JarFile builder
>                    for node in get_all_targets(env):
>                        if(s == str(node)):
>                            target_nodes.append(node)
> 
>this is where 'MANIFEST.MF' as a string lands, and it looks like it doesn't
>match anything in the get_all_targets list, and so is never appended to
>target_nodes. That means it's never there to be picked by jarManifest().

mwichmann said at 2017-12-16 13:05:10
>and this is because of variant dir being in effect:
> 
>XXX not node, checking in
>/home/mats/iotivity.work/out/linux/x86_64/debug/java/examples-java/simpleserver
> 
>in that location, os.path.isfile('MANIFEST.MF') does not succeed; scons knows
>the file is there but Python does not. At the moment I can't suggest a patch, as
>I don't know how to refer back to the source location in the context in which
>the jar.py tool runs (usual tricks that work in an sconscript led to "not found"
>exceptions here)

mwichmann said at 2017-12-17 10:23:30
>if my claim isn't clear, let me add that I believe this is only an issue where
>the setup has been of the form env.VariantDir(build_dir, dir, duplicate=0).  If
>there is no variant dir, or if scons is allowed to duplicate files into the
>variant dir, then os.path.isfile will find a file from the source tree.  I know
>there's something somewhere in the docs that says duplicate=0 is "usually safe"...

dmoody256:
Should this issue be closed now that we merged a supposed fix?

[...]

PR #3098, opened 2018-02-11
Fix Jar to use SCons to check for files instead of python

dmoody256:
This fixes the issue https://github.com/SCons/scons/issues/3042, and adds a test for the case of variant dirs which reflect what is reported in the issue.

## Contributor Checklist:


dmoody256:
This was originally https://github.com/SCons/scons/pull/35, so i resubmitted after Tigris migration

PR #3052, opened 2018-01-18
Add method for individual tools to search for windows install paths which aren't in SCons' defaults + changes to get more tests to pass on windows and/or appveyor

bdbaddog:
Add method for individual tools to search for windows install paths which aren't in SCons' defaults.

- Updated logic for mingw and clang on win32 to search default tool install paths if not
      found in normal SCons PATH.  If the user specifies PATH or tool specific paths they
       will be used and the default paths below will be ignored.
    - Default path for clang/clangxx : C:\Program Files\LLVM\bin
    - Default path for mingw         : c:\MinGW\bin

## Contributor Checklist:

* [ ] I have created a new test or updated the unit tests to cover the new/changed functionality.
* [x] I have updated `master/src/CHANGES.txt` directory (and read the `README.txt` in that directory)
* [ ] I have updated the appropriate documentation

PR #3099, opened 2018-02-11
Windows clang tests

dmoody256:
On windows, a SCons environment will only include the system32 path, unless default windows environment is used in which case the path is setup from the vcvars batch script.

The clang test do not use the msvs environment and therefore do not have the clang or other necessary cygwin binaries in the path when scons goes to build.

This pull request updates the test to add the correct binaries to the path, and skips tests that are meant to use the default environment.

## Contributor Checklist:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,8 @@ image: Visual Studio 2017
 shallow_clone: true
 
 install:
-  - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  #set python version, and cygwin tools for clang tests
+  - "set PATH=%PYTHON%;%PYTHON%\\Scripts;C:\\cygwin64\\bin;C:\\msys64;%PATH%"
   - python --version
   - pip install lxml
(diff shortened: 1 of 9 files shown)


bdbaddog:
Would we better off with more reasonable default installation directories for clang on windows? (which would resolve this without the need for changing the tests?)

dmoody256:
That would probably be the better solution. I wasn't sure the scons way to implement this though.

Would I add some common default paths for clang and cygwin (checking for existence?) or would I look at the current os.environ PATH to find the location of the binaries if they exist, or a combination of both?

bdbaddog:
Changing the logic to have some reasonable default locations would be o.k but importing os.environ's PATH by default is generally not SCons'ian

dmoody256:
Closing this pull request to reopen with the changes in clang tool to use some default paths for clang and cygwin on windows.

bdbaddog:
This is my work in progress for this issue: 
https://github.com/SCons/scons/pull/3052

PR #3101, opened 2018-02-19
Remove forced u" on string value that causes

dragon512:
TypeError: can only join an iterable of bytes (item 3 has type 'unicode'):
...
  File "/home/jkenny/env/lib/python2.7/site-packages/scons-3.0.1/SCons/Action.py", line 185:
    return _function_contents(obj.__call__.__func__)
  File "/home/jkenny/env/lib/python2.7/site-packages/scons-3.0.1/SCons/Action.py", line 301:
    contents = [_code_contents(func.__code__, func.__doc__)]
  File "/home/jkenny/env/lib/python2.7/site-packages/scons-3.0.1/SCons/Action.py", line 264:
    contents.extend(bytearray(',', 'utf-8').join(z))

This is breaking Parts. Not sure why this is not seen in SCons testing as the code that is broken is 
SCons.Action._object_contents(...)

### DIFF
--- a/src/engine/SCons/Action.py
+++ b/src/engine/SCons/Action.py
@@ -678,7 +678,7 @@ class _ActionAction(ActionBase):
                 source = executor.get_all_sources()
             t = ' and '.join(map(str, target))
             l = '\n  '.join(self.presub_lines(env))
-            out = u"Building %s with action:\n  %s\n" % (t, l)
+            out = "Building %s with action:\n  %s\n" % (t, l)
             sys.stdout.write(out)
         cmd = None
         if show and self.strfunction:


bdbaddog:
Have you tried with both py2.7 and py3?

dragon512:
I tried it only on py2.. still porting parts to py3. However, I have code that runs on both py3 and py2 and the scons code seems to be doing this backward. It should be trying to handle everything as normal strings and only be encoding on write to a file or console

bdbaddog:
Doesn't seem right that this code would affect the object contents..  Have you debugged into it and found out what's in z?

dragon512:
I need to look into this more as this is failing on python27 here but working for me on python27.

I did look into Z. it was a list with Unicode 
Z if I recall correctly was the value of u"Building .." form line 681, then some string objects and byte array objects. the complaint about having three string types was correct.

The parts tests fail on it easily enough

bdbaddog:
Strange. A lot of that is new code or mostly rewritten code to deal with py2/3 issues.  
Signatures should always be bytes and not strings. (especially not encoded strings)

Do you have an Action function in Parts which is somehow wrapping and Action object?

bdbaddog:
That's not intended to become part of the signature. That's just output
from the Action object.
It can generate it's own signature I believe.

Are the Actions your comparing CommandAction()'s or some other kind of
action object?
If CommandAction, then get_presig() is likely what you want.

dragon512:
I was calling SCons.Action._object_contents(...)
the input to the  SCons.Action._object_contents function is:
`"${_stripResolveString(__env__, TARGETS, _pdbStripAction)}"`
Z is: (
`[

 bytearray(b"{int:__builtin__}[[[(<type \'object\'>, ()), [(<type \'int\'>, (<type \'object\'>,))]]]]{{}}
{{{denominator=1,imag=0,numerator=0,real=0}}}"),

 ' and ',

 '\n  ',

 u'Building %s with action:\n  %s\n',

 'os.chdir(%s)\n',

 'PRINT_CMD_LINE_FUNC',

 'executor',

 'os.chdir(%s)'

]`

bdbaddog:
What were you passing to _object_contents()... ?

dragon512:
A string of the action.. in this case it is:

"${_stripResolveString(__env__, TARGETS, _pdbStripAction)}"

bdbaddog:
Can you post a URL to the file where you make this call in Parts?

dragon512:
https://bitbucket.org/sconsparts/parts/src/3258a0ff867c5c48faac46993837a05a13d5e474/parts/pnode/part.py?at=master&fileviewer=file-view-default#part.py-1388

however it might help to understand:
https://bitbucket.org/sconsparts/parts/src/3258a0ff867c5c48faac46993837a05a13d5e474/parts/pnode/part.py?at=master&fileviewer=file-view-default#part.py-1413

which is trying to get some env diff values... from this code
https://bitbucket.org/sconsparts/parts/src/3258a0ff867c5c48faac46993837a05a13d5e474/parts/pnode/part.py?at=master&fileviewer=file-view-default#part.py-671

basically i have two different environments and I want to get what is different about it. In this case there is a key that has a value of an action object of some kind, so I call code to get the content from SCons. This call fails. I need to see what this key is that has this object in it...

bdbaddog:
O.k. So instead of using ActionBase, you should have a case for each of the defined action classes and then if none of them check for actionbase, doing something "smart" to determine their differences. In the case of CommandAction, it's basically just the string.  That should be better and faster.

dragon512:
I am open to new idea for this. Why wouldn't SCons know how to get the context better than me. This is data that has to be used to get the sig for scons in the first place. Unless i missed something all these object ( actions, builders, nodes...) have a way to get the content, that is then used to make the sig. In this case I agree the action object is a Command actions. I could "subst" it to get the value but this was being done before. It seems like a bug that this API is returning some of the stuff it is at the moment

dragon512:
OK... now that I have debugged this better, I see this I was probably always doing the wrong thing here. The code in question here which state it is to "Return the signature contents of any Python object." I think is incorrect. It does not handle objects correctly as I look over it again. I think the core mistake is that all Action objects are "callable" but the second case of the _object_contents() function is handling any object that looks like a function, when it should instead treat it like a object.

The more I look at this I am unclear on why this code in SCons is trying to pickle the "function". From what I can understand this "pickled" function is not being loaded again to be called from the "unpicked" state. All this "data" is used for ( correctly me if i am wrong) is to generate sig values to help Scons understand if the context has changed between build. I think this could be handled better.. Anyways I am changing Parts to be smarter about making signatures off of actions.. so i will close this patch.

bdbaddog:
Also... the _ should give you a clue that it probably shouldn't be used
outside the module...

dragon512:
_ is something that "probably" should not be used I agree.. I tend to avoid them.. There are cases in which this is needed. I honestly I am not sure why I was not using get_presig() with a fake target and source

Issue #3102, opened 2018-02-19
JobsTest.py fails intermittantly in the Appveyor and Travis CI environments


dmoody256:
JobsTest.py will intermittently fail in the CI environments. This is believed to be caused by the use of virtual machines used by the different CI's. 

The failure comes from not executing in parallel when it should have:

```
8/292 (2.74%) C:\\Python35\\python.exe -tt src\engine\SCons\JobTests.py
F......
======================================================================
FAIL: runTest (__main__.ParallelTestCase)
test parallel jobs
----------------------------------------------------------------------
Traceback (most recent call last):
  File "src\engine\SCons\JobTests.py", line 255, in runTest
    "the tasks were not executed in parallel")
AssertionError: False is not true : the tasks were not executed in parallel
```

However if the exact same code is run again, then it may pass in the consecutive run. An example of this happening is in this build:
https://travis-ci.org/SCons/scons/jobs/341249882

An example of this failing in Travis is here:
https://travis-ci.org/SCons/scons/jobs/343424806

An example of this failing in Appveyor is here:
https://ci.appveyor.com/project/SCons/scons/build/1.0.38/job/3eikkc3jes5v8s0m

Currently the work around is to retry the test a certain number of times, but this does not always workaround the issue.

dmoody256:
This could be related to the VM's not presenting completely accurate information about the number of cpu cores in the environment. More info here:
https://github.com/travis-ci/travis-ci/issues/4696

bdbaddog:
Does it pass consistantly if you hardcode the # of cpus as 2 in travis?

-Bill

dmoody256:
I haven't tried anything related to hard coding the number of CPUs, but where specifically are you referring to perform this hard-coding?

bdbaddog:
Somewhere here:

https://github.com/SCons/scons/blob/master/src/engine/SCons/JobTests.py#L66

dmoody256:
I'll give it a shot on one of my branches and post the build here

bdbaddog:
might try something like:

if os.environ.get('SCONS_TEST_NUMJOBS',0):
   num_jobs = os.environ['SCONS_TEST_NUMJOBS']

And then specify that in the travis.yml ?

dmoody256:
Number of failures increased with 2 as the num_jobs:
https://travis-ci.org/dmoody256/scons/builds/343653865

This made me curious if the opposite affect is true with increased num_jobs (num_jobs = 1000):
[...]

PR #3104, opened 2018-02-20
JobTests.py parallel issue fix

dmoody256:
This is a fix for https://github.com/SCons/scons/issues/3102, which fixes the issue of intermittent failures on the CI environments.

This uses a different method for checking to see if the tasks are running in parallel or not. It is thought the previously JobTests.py used a method which could render false positives because it only used execution order to determine if the tasks are parallel or not. If the threads execute fast and evenly enough, then it would look like they did not run in parallel.

In a separate test branch, the build below sets the travis script to continuously re-runs the JobTests.py test to try to flush out any intermittent failures:
https://travis-ci.org/dmoody256/scons/builds/343905056

## Contributor Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,7 @@ jobs:
   include:
     - &test_job
       stage: Test
-      script: 
-      # WORKAROUND: attempt to retry JobTests.py if it fails and then continue if it passes, if it fails ten times
-      # then it is a real failure not related to intermittent travis failures
-        - n=0; while [[ $n -lt 10 ]]; do python runtest.py src/engine/SCons/JobTests.py && break; n=$((n+1)); done; if [ "$n" -gt "9" ]; then false; fi
-        - echo "src/engine/SCons/JobTests.py" > exclude_jobtest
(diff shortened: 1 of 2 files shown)


bdbaddog:
Will the test pass if you do less iterations in the sine loop? Maybe just some random amount of 0.01 second sleeps?  6.6 seconds to run the test with your changes

dmoody256:
Removing the sines and only sleeping 0.01 in randomtask takes off 2 seconds to 4.5. Reducing the other sleeps  by a factor or 10 reduces it to 3.5 seconds. Reducing the num_jobs or multiplier for num_tasks can reduce it to less than a second.

I can't tell if the integrity of the test is diminishing with any of those changes. I do worry that if the tasks execute faster than the next thread can start up it would cause the parallel check to fail.

Also there is already a sleep test in this unit test, the random task I believe is testing task which actively use CPU, so I think the sines should stay. I think lowering the total number of sines will retain function and reduce time. Also I think reducing the other sleeps as well will reduce time, and maintain function.

bdbaddog:
Sounds good. Give it a shot.

bdbaddog:
Looks good. merging.

PR #3105, opened 2018-03-11
default shared object suffix on UNIX system should be .so (not .os!)

yanghao:
Looks like GCC implementation fixed this but when using cython/pyext which uses default createCFileBuilders this is not getting fixed alone the path.

This causes when using cython/pyext Tools file to create object file ending with ".os" instead of ".so".

Or if I ".os" has ever being a valid shared object output suffix which I am completely not aware of ...

### DIFF
--- a/src/engine/SCons/Tool/cc.py
+++ b/src/engine/SCons/Tool/cc.py
@@ -90,7 +90,7 @@ def generate(env):
     env['CPPDEFSUFFIX']  = ''
     env['INCPREFIX']  = '-I'
     env['INCSUFFIX']  = ''
-    env['SHOBJSUFFIX'] = '.os'
+    env['SHOBJSUFFIX'] = '.so'
     env['STATIC_AND_SHARED_OBJECTS_ARE_THE_SAME'] = 0
 
     env['CFILESUFFIX'] = '.c'
--- a/src/engine/SCons/Tool/cxx.py
+++ b/src/engine/SCons/Tool/cxx.py
@@ -84,7 +84,7 @@ def generate(env):
     env['CPPDEFSUFFIX']  = ''
     env['INCPREFIX']  = '-I'
     env['INCSUFFIX']  = ''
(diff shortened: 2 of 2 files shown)


bdbaddog:
SHOBJSUFFIX is for the object output from compiling, not for a shared library and thus the usage of .os and not .so

I'm looking at your: PythonObject builder and there's no action.
What are you trying to do?

Ideally you'd bring this discussion the scons users mailing list.

yanghao:
https://stackoverflow.com/questions/22560662/os-file-unrecognized-while-building-glibc

OK, I found one reference out there and this is the first time I know a object file can be named ".os" instead of just ".o".

Yes I got it wrong, thanks.

bdbaddog:
Bring your issue to the scons users mailing list and perhaps someone there can help you sort it out?

yanghao:
By the way, the PYEXTCC builder builds the .c file into .os file, which is then linked into a shared library (.so) file. I override now the .os to ".o" which is making more sense for me.

garyo:
The reason for using ".os" for SharedObject is to prevent accidentally linking /fPIC object files with non-pic, which would cause runtime errors. If you know that's not going to happen in your build, you can just set SHOBJSUFFIX to ".o".

yanghao:
Thanks Gary.

This explains the motivation! I will keep using ".os" instead.

Also found a detailed explanation on PIE/PIC and when to use what, hope
this could help someone else:
https://lists.debian.org/debian-devel/2016/05/msg00309.html

For people like me come from embedded world, I almost always use no-pie
exclusively because the tiny extra performance always beats the relocation
needs.

bdbaddog:
Indeed I had a client with some software in a HPC space. And shared library builds cost them 5-10% performance.  So devel builds were shared, but release were static.

yanghao:
Interesting. in HPC definitely it worth it even more than embedded world.

Issue #3039, opened 2018-01-02
Allow for use of SConstruct.py where SConstruct is not detected


bdbaddog:
This issue was originally created at: 2017-05-07 14:46:20.
This issue was reported by: `garlicbread`.
garlicbread said at 2017-05-07 14:46:20
>Hi,
>One thing I'd like to ask for as a feature request
>is if Scons does not detect the presence of a SConstruct file
>that it also tries searching for a SConstruct.py file
> 
>Typically I've been using Visual Studio with python tools to debug the build 
>scripts - http://grbd.github.io/posts/2016/07/27/scons-builds-with-visual-
>studio/
> 
>However because the file doesn't have an extension this makes it difficult 
>to associate with the Python editor in the IDE.
>One work around I've been using is to create a SConstruct file with the 
>following line in as a workaround
>execfile('SConstruct.py')
> 
>It doesn't need to be a .py extension, it could be something like .scons
>As long as there's some form of extension this should make it easier to 
>associate with python inside IDE's

bdbaddog:
Fixed in next post 3.0.1 release.

PR #3110, opened 2018-04-01
Add SConstruct.py to the search list

Hecatron:
This is a small change to allow scons to pick up on SConstruct.py if SConstruct isn't found
in relation to https://github.com/SCons/scons/issues/3039

At the moment to get debugging to work within Visual Studio I have to create a dummy SConstruct file within the following in
```
exec(compile(open('SConstruct.py').read(), 'SConstruct.py', 'exec'))
```
Then have the real code located within SConstruct.py
This is because Visual Studio doesn't recognize the code within the file as python due to the lack of a file extension. This just makes life a little bit easier.
SConscript files within sub-directories are fine since you need to specify those with a file path within the script sitting above anyway.

### DIFF
--- a/doc/design/native.xml
+++ b/doc/design/native.xml
@@ -64,8 +64,9 @@
  <para>
 
   By default, the &SCons; utility searches for a file named
-  &SConstruct;, &Sconstruct;, &sconstruct;, &SConstruct.py or &sconstruct.py; (in that order) in the
-  current directory, and reads its configuration from the first file
+  &SConstruct;, &Sconstruct;, &sconstruct;, &SConstruct.py, &Sconstruct.py
+  or &sconstruct.py; (in that order) in the current directory,
+  and reads its configuration from the first file
   found.  A <option>-f</option> command-line option exists to read a
   different file name.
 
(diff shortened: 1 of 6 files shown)


bdbaddog:
Can you add a test? An likely will need an update to at least the manpage.

Hecatron:
I can if you want, I wasn't sure if it'd need one

bdbaddog:
To be consistent with the SConstruct filenaming options can you also add sconstruct.py (all lower)

Hecatron:
If using the .py extension is a problem (if your worried about folks trying to run it directly)
then we could use something else like SConstript.scons as long as it's possible to use a file with an extension then it should possible to set the IDE to interpret it as a python file

bdbaddog:
.py is fine. We've had similar requests before.
If you can add the lowercase to be consistant and update the changed docs, and add a simple test, we should be good to merge.

Hecatron:
those changes should now already be in (for lowercase / to the docs / tests etc)
the new search order is now ['SConstruct', 'Sconstruct', 'sconstruct', 'SConstruct.py', 'sconstruct.py']
the last 2 are the ones I've added in, if you also want Sconstruct.py let me know and I'll add that in as well

bdbaddog:
May as well be consistant. Please go ahead and add Sconstruct.py.
Thanks!

Hecatron:
Okay now done

PR #3111, opened 2018-04-04
Fix EnvironmentError printing bug

moroten:
EnvironmentErrors does not need to have the attributes strerror or
errno. This commit makes the code robust for such situation.

One example of such EnvironmentError is CacheDir.py:
raise SCons.Errors.EnvironmentError(msg)

Signed-off-by: Fredrik Medley <fredrik.medley@autoliv.com>

## Contributor Checklist:


bdbaddog:
Can you explain the need for this patch a bit, or even better bring the discussion to the users or developers mailing list?

bdbaddog:
Can you update the CHANGES.txt blurb to indicate in what case this was failing that won't fail now?

bdbaddog:
Ok. Looks good. Merging.

PR #3114, opened 2018-04-05
Use a lock file when initializing the cache

moroten:
There is a documented race condition when initializing the cache.
The race condition exists both between processes and threads within one
process, the latter is undocumented but more likely. Add a lock file so
that only one thread in one process at a time is allowed to initialize
a SCons cache.

Note that the implementation should work also for multiple computers
initializing a cache directory on a network drive, although that has not
been tested.

Signed-off-by: Fredrik Medley <fredrik.medley@autoliv.com>

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/CacheDir.py
+++ b/src/engine/SCons/CacheDir.py
@@ -32,6 +32,7 @@ import json
 import os
 import stat
 import sys
+import time
 
 import SCons.Action
 import SCons.Warnings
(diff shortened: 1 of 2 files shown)


bdbaddog:
Are you seeing this problem in practice?

moroten:
Yes. Our automatic build environment clears the workspace, where the cache is configured to reside, every day. This race condition is basically triggered at least once every day and failure is reported to the user.

The "problem" is that we let SCons run multiple threads and the following happens:
1. The json config file is opened for writing by thread A.
2. Thread B reads the empty file, fails parsing it and reports an error.
3. Thread A writes the content to the config file and closes it.

Note that the next user of the build workspace have no problem as the config file is written correctly, so no cleanup is needed, just a rerun.

bdbaddog:
So you're saying that a blank cachedir and running scons -j will  cause
this problem to happen?

moroten:
Yes, in our environment. But it is a race condition so it is not for sure it happens for everyone. Does e.g. anti-virus software affects what happens?

bdbaddog:
Anti-virus on pc's definitely can affect the speed you write files to the
disk.
So it could come into play.

src/engine/SCons/CacheDir.py
```diff
@@ -145,56 +148,91 @@ def __init__(self, path):
...
+
+        cache_create_tries = 0
+        while not os.path.exists(config_file):
+            # Lock the whole process of creating the directory and config file.
+            try:
+                os.remove(lock_file)
+            except OSError:
+                pass
+            try:
+                lock = os.open(lock_file, os.O_CREAT | os.O_EXCL)
```
mwichmann: Relatively convinced this is not going to work. On Linux/UNIX  os.remove will remove the directory entry for a file, even if the file remains open.  Then the open + create + exclusive will complete happily, because open works off the directory entry. So now you have two processes thinking they hold the lock file.

mwichmann:
It's possible the new code will work, haven't looked yet.  Locking is hard, locking without killing performance is harder.  There are existing implementations in Python which it might be safer to take advantage of... maybe scons should use the multiprocessing module which already has this worked out? aka  `from multiprocessing import Process, Lock`

bdbaddog:
@moroten  - If you changed the code I don't see the commit in the PR yet. Perhaps you haven't pushed your branch to github?

moroten:
@bdbaddog - The PR has been updated. The old one was https://github.com/SCons/scons/commit/e3e414c1256faf1f0854e3f844c0216ce7df73a6 and the new version, which has been rebased as well, is https://github.com/SCons/scons/pull/3114/commits/70a3d6a3141bf49fafeadaaccbf15a5dae23bcd0. I did amend the commit followed by force push, but maybe you did expect a second commit with only the fix of the Linux case.

@mwichmann - When reading the original code and comments, it stated that one goal was to keep multiple processes out of the race hazard, not just handling the internal threads during a parallel build which would easily been solved by `threading.Lock`. What I understand, by quickly reading the documentation, `multiprocessing` is a way of forking a running Python process, not to handle independently spawned processes.

bdbaddog:
Seems like an over complicated solution. Rather than fight with multple threads in the same scons process, why not just initialize before the taskmaster starts running and/or when the cachedir argument(s) are processed.
And do it via write to x.tmp, mv x.tmp x type solution?

mwichmann:
What is the cache config file for, anyway?  Seems like it can contain only one key, which is how many characters of a signature to use as a subdirectory name. And that's always two, unless some old cache exists that didn't use two... which at this stage seems quite unlkely. How many years would someone keep an old cachedir around? So it all appears to be a transition thing that maybe we don't even need to keep? (yes I see an external tool can reconfigure it)

bdbaddog:
No the intent is not that it would be transitional, but rather persist and
help with migrations going forward and/or store more info if needed.
Also if you switch from an older version which didn't have the file, scons
will detect..

mwichmann:
detect... unless you ran into a race where the cachedir is found but the config file hasn't been created yet :)   anyway I feel the mechanism is overly complex for how little it does, but you've answered the intent, so fine.  In general this sort of thing wants a reader/writer lock: all can get read access, unless the writer is writing; the writer can't get the write lock if someone is reading but should be weighted in its favor so it doesn't have to wait too long. That seems like overkill for such a simple config file that is unlikely to be changing in normal operation - the only way I can see for the value to change is the scons-configure-cache program, which is external to scons and thus wouldn't be participating in any locks anyway.  So I think the suggestion of "get this done early, before there are multiple tasks going" is a smart one.

bdbaddog:
Is this fixed by changed which are in 3.1.1 to resolve cachedir races?
@mwichmann @chasinglogic

mryan1539:
I'm still seeing this issue in 3.1.2.  I temporarily worked around it by adding a `env.get_CacheDir()`  call in my main SConstruct right after calling `CacheDir()` and creating my root environment.  At first glance, it seems like having `CacheDir()`  be responsible for the initialization of the cache directory might be appropriate, perhaps?

bdbaddog:
> I'm still seeing this issue in 3.1.2. I temporarily worked around it by adding a `env.get_CacheDir()` call in my main SConstruct right after calling `CacheDir()` and creating my root environment. At first glance, it seems like having `CacheDir()` be responsible for the initialization of the cache directory might be appropriate, perhaps?

Sorry for the long delay in responding. 
Yes that sounds like it would do what I'd suggested above to open the cachedir lock file before the threads are spawned.
Can you update the patch to do that?

bdbaddog:
Closing this PR in favor of #3864

bdbaddog:
@moroten - any chance you can try the logic in #3864 ?

moroten:
We are not using Scons any more, so I am unable to try the logic. I remember that just running Scons in a for-loop in bash for 100 times on a simple build easily triggered the problem several times.

bdbaddog:
@moroten - out of curiosity what did you switch to and what were the reasons?

moroten:
> @moroten - out of curiosity what did you switch to and what were the reasons?

Before, we had multiple build systems of which one was Scons. Our investigations a few years ago showed that porting it all to Scons would result in startup times, i.e. reading the build description before starting running any compile action, would take over a minute. Developers would be frustrated with such times. We decided to go with Bazel instead, as it performed the same incremental code+test loop in a few seconds.

PR #3116, opened 2018-04-11
Fix for issue #3013 (and duplicate 3018). Side effects cause internal error with variant_dir

### LINKED ISSUE #3013: Side effects cause internal error with variant_dir
This issue was originally created at: 2015-11-12 06:53:27.
This issue was reported by: `sross`.
sross said at 2015-11-12 06:53:27
>When including a SideEffect statement in a SConscript built with a variant_dir, 
>there is an internal error, e.g.:
> 
>>> scons 
>scons: Reading SConscript files ...
>scons: done reading SConscript files.
>scons: Building targets ...
>scons: building associated VariantDir targets: build
>scons: done building targets.
> 
>scons: *** Found dependency cycle(s):
>  Internal Error: no cycle found for node build (<SCons.Node.FS.Dir object at 
>0x20da620>) in state pending
>  build/output -> build/output
> 
>File "/usr/local/lib/scons-2.4.0/SCons/Taskmaster.py", line 1043, in cleanup

sross said at 2015-11-12 06:55:17
>Created an attachment (id=962)
>Demonstration of issue

sross said at 2015-11-12 06:56:21
>See the attached files for a demonstration of the problem...running scons 
>produces the error.

sross said at 2015-11-12 07:02:56
>Created an attachment (id=963)
>Demonstration of issue with parallel build

sross said at 2015-11-12 07:05:53
>Sorry, it looks like my first example was showing something slightly different 
>(but still a bug, it reports internal error).  Please see the second example, 
>fails_with_parallel_build.tar.gz.
> 
>The second example passed with scons, fails with scons -j2.

bernardb said at 2016-02-09 08:39:48
>I've also hit this same bug, and I think I've tracked it down to an error in Taskmaster's 
>logic around the handling of side-effects. A side-effect may cause its parent to be added 
>to the waiting_parents list, which increments the parent's ref_count.
> 
>The reference count for side-effects is only decremented by postprocess() if the side-
>effect node is in the state NODE_EXECUTING. However, in some cases, 
>executed_without_callbacks/executed_with_callbacks is called before postprocess(), which 
>sets the state of the side-effect nodes to NODE_NO_STATE. I confess I don't understand the 
>logic here, but I think the issue is that when postprocess() gets called, the reference 
>count does not get decremented because the side-effect nodes are not in the NODE_EXECUTING 
>state.
> 
>So my first attempt was to modify postprocess() to also iterate over a side-effect's 
>waiting_parents if it was in the state NODE_NO_STATE. This gave rise to a second issue 
>which is that side-effects may get generated more than once, and the waiting_parents set 
>was never cleared, nor were the nodes removed from pending_children. Fixing both of these 
>seems to resolve the issue for me, and also fixes the test case Stefan has attached.
> 
>Patch is attached. I'm not hugely confident in this, so please take this as a stab in the 
>right direction. I've tested the existing unit tests and it seems to pass. I don't know how 
>or have time to turn Stefan's test case into a unit test, but I hope some developer can 
>help there!

bernardb said at 2016-02-09 08:41:09
>Created an attachment (id=964)
>First stab at fixing this bug

bdbaddog said at 2016-02-09 18:05:58
>Could we ask you to submit a pull request on bitbucket?
> 
>https://bitbucket.org/scons/scons/wiki/DeveloperGuide/MercurialWorkflows
> 

bernardb said at 2016-02-09 20:45:41
>I've created a pull request at https://bitbucket.org/scons/scons/pull-requests/303/fix-handling-of-side-effects-in-taskmaster/diff

Votes for this issue: 11.

sross attached [side_effect_error.tar.gz](https://github.com/SCons/tigris-issue-attachments/blob/master/3013/962/side_effect_error.tar.gz) at 2015-11-12 06:55:17.
>Demonstration of issue

sross attached [fails_with_parallel_build.tar.gz](https://github.com/SCons/tigris-issue-attachments/blob/master/3013/963/fails_with_parallel_build.tar.gz) at 2015-11-12 07:02:56.
>Demonstration of issue with parallel build

bernardb attached [scons-3013-fix.diff](https://github.com/SCons/tigris-issue-attachments/blob/master/3013/964/scons-3013-fix.diff) at 2016-02-09 08:41:09.
>First stab at fixing this bug

### LINKED ISSUE #3018: scons -j num causes dependency cycle, no cycle found for node
This issue was originally created at: 2015-12-29 04:13:46.
This issue was reported by: `anticimex`.
anticimex said at 2015-12-29 04:13:46
>Referring to the scons-users thread at 
>https://pairlist4.pair.net/pipermail/scons-users/2015-November/004224.html , 
>there is a bug with parallel builds, provoked by all of variant_dir/SideEffect/parallel.
> 
>The bug:
```
$ scons -j 8
scons: Building targets ...
scons: building associated VariantDir targets: build
make_file(["build/output"], ["test.cpp"])
scons: done building targets.
 
scons: *** Found dependency cycle(s):
  Internal Error: no cycle found for node build (<SCons.Node.FS.Dir object at 
0x25eb210>) in state pending
 
File "/usr/lib/scons/SCons/Taskmaster.py", line 1026, in cleanup
```

> 
>The test code.
>SConstruct:
>==================
```
env = Environment()
 
def make_file(target, source, env):
    with open(str(target[0]), 'w') as f:
        f.write('gobldygook')
    with open(str(target[0]) + '_side_effect', 'w') as side_effect:
        side_effect.write('anything')
 
env.Append(BUILDERS =
    {'make_file': Builder(action=Action(make_file))})
 
SConscript('SConscript',
     variant_dir = 'build',
     exports={'env':env},
     duplicate=0)
```
>==================
> 
>SConscript:
>==================
```
Import('env')
 
primary = env.make_file('output', 'test.cpp')
this_causes_problems = env.SideEffect('output_side_effect', 'output')
```
>==================
> 
>test.cpp:
>==================
```
#include <iostream.h>
 
main()
{
    cout << "Hello World!";
    return 0;
}
```
>==================
> 
> 
>In our case, we're trying to build a project with ~200 SConscript files that 
>fails like this bug, when doing parallel build, and serial build just takes so 
>long (>45m).
>Our dev cycles would improve severely if we were able to properly do parallel 
>builds in the project.

Votes for this issue: 2.

bdbaddog:
From bitbucket PR: https://bitbucket.org/scons/scons/pull-requests/303

Fixes #3013 
Fixes #3018 

Mixing SideEffects, VariantDir and parallel build could lead to stack trace due to uncleared waiting parents on sideeffect nodes.

Original patch by @bblackham

## Contributor Checklist:

* [x] I have updated the appropriate documentation - None

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -7,6 +7,9 @@
 
 RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
 
+  From Bernard Blackham:
+   - Fixed handling of side-effects in task master (fixes #3013).
+
   From Daniel Moody:
     - Set the pickling protocal back to highest which was causing issues 
       with variant dir tests. This will cause issues if reading sconsigns
--- a/src/engine/SCons/Taskmaster.py
+++ b/src/engine/SCons/Taskmaster.py
@@ -470,14 +470,23 @@ class Task(object):
                 pending_children.discard(t)
             for p in t.waiting_parents:
                 parents[p] = parents.get(p, 0) + 1
+            t.waiting_parents = set()
 
         for t in targets:
             if t.side_effects is not None:
                 for s in t.side_effects:
                     if s.get_state() == NODE_EXECUTING:
                         s.set_state(NODE_NO_STATE)
+                        
+                    # The side-effects may have been transferred to
+                    # NODE_NO_STATE by executed_with{,out}_callbacks, but was
+                    # not taken out of the waiting parents/pending children
+                    # data structures. Check for that now.
+                    if s.get_state() == NODE_NO_STATE and s.waiting_parents:
+                        pending_children.discard(s)
                         for p in s.waiting_parents:
                             parents[p] = parents.get(p, 0) + 1
+                        s.waiting_parents = set()
                     for p in s.waiting_s_e:
                         if p.ref_count == 0:
                             self.tm.candidates.append(p)
--- a/test/SideEffect/Issues/3013/files/SConscript
+++ b/test/SideEffect/Issues/3013/files/SConscript
@@ -1,0 +1,5 @@
+Import('env')
+
+primary = env.make_file('output', 'test.cpp')
+this_causes_problems = env.SideEffect('output_side_effect', 'output')
+
--- a/test/SideEffect/Issues/3013/files/SConstruct
+++ b/test/SideEffect/Issues/3013/files/SConstruct
@@ -1,0 +1,21 @@
+env = Environment()
+
+def make_file(target, source, env):
+    with open(str(target[0]), 'w') as f:
+        f.write('gobldygook')
+    with open(str(target[0]) + '_side_effect', 'w') as side_effect:
+        side_effect.write('anything')
+
+env.Append(
+    BUILDERS={'make_file': Builder(action=Action(make_file))}
+)
+
+env.objdir = 'build'
+
+SConscript(
+    'SConscript',
+    variant_dir=env.objdir,
+    exports={'env':env},
+    duplicate=0
+)
+
--- a/test/SideEffect/Issues/3013/files/test.cpp
+++ b/test/SideEffect/Issues/3013/files/test.cpp
@@ -1,0 +1,2 @@
+void some_function() {}
+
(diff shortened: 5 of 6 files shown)


garyo (approved):
LGTM. The Taskmaster code you're updating is surprisingly dense; maybe since you're in there, add a couple of comments (on the existing code) to help future devs?

bdbaddog:
Gonna go ahead and merge. If I get some time I'll do another pull request with just comments.

PR #3122, opened 2018-04-30
Get rid of QMTest dir and move files to testing/framework. QMtest hasn't been used in years

bdbaddog:
A bit of cleanup in the source tree.

No need to confusingly have a directory named QMtest when we've not used it since 2.3.2 (2014)

Once this merges, I'll update the relevant wiki and any other docs.
Updating those now would be premature and could cause confusion.

## Contributor Checklist:


garyo:
LGTM!

bdbaddog:
Russel,

Can always back it out.
Was looking for a sanity check from any other experienced SCons dev.

Do you concur with the change?

-Bill

russel:
All the tests ran for me as expected,so I see no reason not to exactly because QMTest is long gone.

PR #3127, opened 2018-05-20
Hashable literal

maiphi:
Make class SCons.Subst.Literal hashable.

When using Python 3, substitution of Literal objects requires the objects to be hashable, otherwise an error will be thrown. The returned hash value is that of the lstr member.

This is not a feature update so I didn't write a test case and there is no need for updated documentation/readme imo.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -7,6 +7,10 @@
 
 RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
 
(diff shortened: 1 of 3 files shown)


bdbaddog:
Can you add a test case with your simple reproducer?
We're missing a testcase for this or we would have caught it prior to 3.0.0 release.

maiphi:
Added a test which will hopefully work on all platforms (not using $_RPATH).

garyo:
LGTM!

PR #3132, opened 2018-05-26
Added missing FORTRANMODDIRPREFIX to the gfortran tool.

maiphi:
Added the missing FORTRANMODDIRPREFIX=-J to the gfortran tool. Without this, FORTRANMODDIR cannot be used.

Also added a test which compiles a .f90 source file to an executable with gfortran and checks that module files are place in FORTRANMODDIR correctly, the executable is created and the build is recognised as up-to-date (i.e. the dependency scanner finds the module file in the module directory).

Documentation update is not required, because this restores the documented behaviour.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -10,6 +10,7 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
   From Philipp Maierhöfer
     - Added a __hash__ method to the class Scons.Subst.Literal. Required when substituting Literal
       objects when SCons runs with Python 3.
+    - Added missing FORTRANMODDIRPREFIX to the gfortran tool.
 
   From Richard West:
     - Add SConstruct.py, Sconstruct.py, sconstruct.py to the search path for the root SConstruct file.
--- a/test/Fortran/gfortran.py
+++ b/test/Fortran/gfortran.py
@@ -1,14 +1,4 @@
-"""SCons.Tool.gfortran
-
-Tool-specific initialization for gfortran, the GNU Fortran 95/Fortran
-2003 compiler.
-
-There normally shouldn't be any need to import this module directly.
-It will usually be imported through the generic SCons.Tool.Tool()
-selection method.
-
-"""
-
+#!/usr/bin/env python
 #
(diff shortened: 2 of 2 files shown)


bdbaddog:
What happens when FORTRANMODDIR is not specified?

maiphi:
The module directory must be passed to the compiler with the option `-J<moddir>`. If FORTRANMODDIRPREFIX is empty (without this patch), the "-J" is missing and `<moddir>` is interpreted as an input file:
```
$ gfortran -o test1.o -c -Imodules modules test1.f90
gfortran: warning: modules: linker input file unused because linking not done
```
The source file is compiled, but the module file is placed in the working directory instead of the module directory (the directory will be created and remain empty). The dependency scanner expects the module file in the module directory, though. Therefore the build will never be up-to-date and `scons -c` will not remove the module file.
The correct command (which we get with FORTRANMODDIRPREFIX=-J set) is
`$ gfortran -o test1.o -c -Imodules -Jmodules test1.f90`

bdbaddog:
Yes. I understand why you made the change.
I'm just wondering what if FORTRANMODDIR is not specified?
What's the effect.
Does it cause an error now that FORTRANMODDIRPREFIX is set?

maiphi:
Sorry, misread your question. When FORTRANMODDIR is not set, everything works as before. FORTRANMODDIRPREFIX is prepended to every directory in FORTRANMODDIR and added to the build command. If no directory is specified in FORTRANMODDIR, there is nothing to prefix with "-J" and no such option will end up on the command line (it works exactly like include dirs):
`env = Environment(FORTRANMODDIRPREFIX='-J')`
`env.Program('test1', 'test1.f90')`
-->
`gfortran -o test1.o -c test1.f90`
`gfortran -o test1 test1.o`

If you look at the ifort tool, you'll see FORTRANMODDIRPREFIX="-modules ", which for ifort does the same as "-J" does for gfortran. It seems that the one who wrote the gfortran tool just forgot it. I.e. the FORTRANMODDIR feature is currently (and has always been) broken in the gfortran tool.

bdbaddog:
Test failure on travis ci.
Can you take a look and resolve?

PR #3133, opened 2018-05-27
Make test FORTRANMODDIR.py work with non-empty FORTRANMODDIRPREFIX.

maiphi:
Fixed the test FORTRANMODDIR.py which failes since FORTRANMODDIRPREFIX has been set in the gfortran tool.

Diagnosis: SCons is working correctly here, but the test expects wrong results.

I fixed it in such a way that it comes as closely as possible to what I think was the intention of the original author. It verifies the following:
* _FORTRANMODFLAG is correctly constructed from FORTRANMODDIRPREFIX and FORTRANMODDIR.
* The dependency scanner does not expect a module file to be created from a "module procedure" statement.
* The dependency scanner expects the module files to be created in the correct module directory as set with FORTRANMODDIR (this is verified by the test.up_to_date()).

Note that the original Fortran source file contained invalid code. Doesn't really matter here, because the test is synthetic (not calling a compiler, but a Python script instead). Anyway, IMO the test should act on valid code, so I fixed it.

## Contributor Checklist:

### DIFF
--- a/test/Fortran/FORTRANMODDIR.py
+++ b/test/Fortran/FORTRANMODDIR.py
@@ -24,6 +24,16 @@
 
 __revision__ = "__FILE__ __REVISION__ __DATE__ __DEVELOPER__"
 
+"""
+Verify the following things:
+* _FORTRANMODFLAG is correctly constructed from FORTRANMODDIRPREFIX and
+  FORTRANMODDIR.
+* The dependency scanner does not expect a module file to be created
+  from a "module procedure" statement.
+* The dependency scanner expects the module files to be created in the correct
+  module directory (is is verified by the test.up_to_date()).
+"""
+
 import TestSCons
 
 _python_ = TestSCons._python_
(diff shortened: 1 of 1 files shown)


test/Fortran/FORTRANMODDIR.py
```diff
@@ -24,68 +24,76 @@
...
-         END SUBROUTINE P
-      END MODULE MOD_FOO
-      MODULE PROCEDURE MOD_BAR
-         IMPLICIT NONE
-         CONTAINS
-         SUBROUTINE P
-            PRINT *,'mod_bar'
-         END SUBROUTINE P
-      END MODULE MOD_BAR
+module mod_foo
```
bdbaddog: Is this syntax specific (non compatible with older) to a particular fortran version?

bdbaddog:
Looks good. Thanks for clarifying.
It's been a while since I did any Fortran.. :)

PR #3134, opened 2018-05-28
Set the pip installation directory to be non version specific by default

Hecatron:
Hi,
I've recently changed setup.py so that by default pip install installs to a non version specific directory
such as
`pythondir\Lib\site-packages\scons`
instead of the original
`pythondir\Lib\site-packages\scons-versionnumber`

This makes life a bit easier when writing external tools, where your using pytest and need to know where the installed scons directory is


bdbaddog:
I'll give it a try with SCons' own build logic. Not currently exercised by CI. (yet)

bdbaddog:
Looks good. merging.

PR #3140, opened 2018-06-23
Do not reuse tempfiles

simontegelid:
Actions lists where each action use `TEMPFILE` cannot use the same tempfiles. Therefore we cannot store the `tempfile_cmdlist` in neither the target nor the source attributes since those are common for all actions.

Reusing the same tempfile causes the same command to be invoked for all actions in the list. However, not reusing tempfiles will create unneeded tempfiles on `strfunction` invocation. This is better than producing incorrect results, but leaves the tempfile in a bad state.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -65,6 +65,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Accommodate VS 2017 Express - it's got a more liberal license then VS
       Community, so some people prefer it (from 2019, no more Express)
     - vswhere call should also now work even if programs aren't on the C: drive.
+    - Add an alternate warning message cl.exe is not found and msvc config
+      cache is in use (SCONS_CACHE_MSVC_CONFIG was given) - config cache
+      may be out of date.
 
 
 RELEASE 3.1.2 - Mon, 17 Dec 2019 02:06:27 +0000
--- a/src/engine/SCons/Tool/MSCommon/vc.py
+++ b/src/engine/SCons/Tool/MSCommon/vc.py
@@ -22,6 +22,9 @@
 #
 
 # TODO:
+#   * gather all the information from a single vswhere call instead
+#     of calling repeatedly (use json format?)
+#   * support passing/setting location for vswhere in env.
 #   * supported arch for versions: for old versions of batch file without
 #     argument, giving bogus argument cannot be detected, so we have to hardcode
 #     this here
@@ -41,7 +44,6 @@ import subprocess
 import os
 import platform
 import sys
-from contextlib import suppress
 from string import digits as string_digits
 from subprocess import PIPE
 #TODO: Python 2 cleanup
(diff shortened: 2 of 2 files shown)


bdbaddog:
Did you mean ".. in either the target nor the source attributes since those are common for all actions."
and not "neither" ?

bdbaddog:
Can you provide an example SConstruct to demonstrate the problem?
Ideally discuss this in the users mailing list?

simontegelid:
If the `tempfile_cmdlist` is stored as an attribute somwhere in the targets (or sources), all the subsequent actions in the same action list that is also using `TEMPFILE` will get the tempfile from the first `TEMPFILE` invocation. I provided a testcase in the pull request that exemplifies the problem, which is why I made a pull request and not opened a discussion in the users mailing list.

bdbaddog:
Bringing it to the users mailing list allows more eyes on the issue and potentially understanding why the code is as it is.
As you've removed some code added to fix some issues.
It wouldn't have been added arbitrarily.

Here's the email regarding the previous changes which you've altered.
https://pairlist2.pair.net/pipermail/scons-dev/2014-November/002041.html

Are those issues re-broken by your changes?

simontegelid:
Yes, as I mentioned "not reusing tempfiles will create unneeded tempfiles on strfunction invocation" but the current solution is producing incorrect results. I will bring it to the users mailing list for discussion.

simontegelid:
The users mailing list didn't get much attention. The solution for the duplicate tempfiles are not correct imo, since it gives incorrect builds for the test case I added. Any ideas how we can solve that in another way?

bdbaddog:
Would:

> Actions lists with more than one action using ${TEMPFILE("TEMP FILE CONTENT")} should not use the
> same tempfile for different "TEMP FILE CONTENT"

Be a better description than:

> Actions lists where each action use TEMPFILE cannot use the
> same tempfiles.

Also note as is the test fails against current master branch with missing xxx.py and yyy.py.

Also. I don't think the test captures checking that the tempfile for each action has a different name?

Seems like an optimal solution would have a single temp file for each item in an action list?
(And not a temp file per usage of each action?)

bdbaddog:
This should be superseded by merged PR #3543

simontegelid:
Oh, sorry for leaving this behind for a long time. I just rebased it and took a look at the added test. I don't really get the details of how the tests are being run, I just used `test/TEMPFILEPREFIX.py` as a base for my test.

Running `python3 ./runtest.py test/TEMPFILE-actionlist.py` gives

```
1/1 (100.00%) /usr/bin/python3 test/TEMPFILE-actionlist.py
PASSED
```

What am I missing? Your proposed test seems more suitable, but I just fail to run it...

simontegelid:
The PR was apparently closed when I force-pushed my rebase..? That wasn't my intention.

Edit: Pushed the wrong ref... Fixed now.

PR #3155, opened 2018-07-22
Fix some tests

dragon512:
Fix some tests that are failing
Add a missing slash to test on test/Execute.py so it matches correctly

This fixes for me:
test\Execute.py
test\MSVC\msvc.py

## Contributor Checklist:


src/engine/SCons/Node/__init__.py
```diff
@@ -637,8 +637,10 @@ def get_executor(self, create=1):
         """Fetch the action executor for this node.  Create one if
         there isn't already one, and requested to do so."""
         try:
-            executor = self.executor
-        except AttributeError:
+            if self.executor is None:
```
bdbaddog: It looks like the exception won't be thrown based on your previous comments about slots usage. Additionally I think it should only create the Null executor if create is True?

bdbaddog:
What python/OS/MSVC version(s) are you seeing the crash(es) on?

src/engine/SCons/Node/__init__.py
dragon512: I agree with the create argument.
this fixes the msvc test. I had some code that this also fixes. .. let me tweak.

dragon512:
The executor test was failing with some stuff in Parts I was trying to add ( prototype stuff)
it was also casing the test\MSVC\msvc.py to fail.

I corrected the logic. based on your observation. The new logic should be more correct and only try to create the executor when create is True now

bdbaddog:
We need a test which fails without parts without your changes?

test/Execute.py
```diff
@@ -86,7 +86,7 @@
 if sys.platform == 'win32' and sys.version_info[0] == 2:
     expect = r"""scons: \*\*\* Error 1
 scons: \*\*\* Error 2
-scons: \*\*\* nonexistent.in/\*\.\*: (The system cannot find the path specified|Das System kann den angegebenen Pfad nicht finden)"""
+scons: \*\*\* nonexistent.in\\\*\.\*: (The system cannot find the path specified|Das System kann den angegebenen Pfad nicht finden)"""
```
mwichmann: I can confirm this change does fix the Execute.py fail  for Python 2.7. Be nice to get it in, even if the rest of the patch is held up. I'm still not clear on why scons provides different error messages in this case (python2 vs. python3) - is that error message from python itself?

dragon512: I might separate the patch to remove the stuff I have not had time to make a good test case for as of yet. Logically I know it is a good fix. I need to reproduce the case again in which I had this crash.

bdbaddog:
Looks like this pull requests is for multiple different things?

Can you break them up into different pull requests?

When do you see test/Execute.py failing (without your code changes)

mwichmann:
It fails on any test run on Windows with Python 2.7. You can see it any any AppVeyor build under this label:
`Environment: PYTHON=C:\Python27, BUILD_JOB_NUM=2`

bdbaddog:
Does the change to node/__init__.py fix the test/execute.py?
Or just the change in that test?

mwichmann:
For `Execute.py`, it's just string matching in the test that got updated, the pattern in there now for py2 doesn't work. 
And as I noted above, I don't quite get why scons is emitting different messages between py2 and py3 in this case. the py3 case looks like the one we'd expect. "The system cannot find..." is the Windows part. but what comes before it is all scons afaict. Py2:
```
    Copy("nonexistent.out", "nonexistent.in")
    scons: *** nonexistent.in\*.*: The system cannot find the path specified
```
Py3:
```
    Copy("nonexistent.out", "nonexistent.in")
    scons: *** nonexistent.in: The system cannot find the path specified
```

bdbaddog:
test\Execute.py Passed here:
http://buildbot.scons.org/#/builders/8/builds/30/steps/1/logs/stdio

bdbaddog:
Check out a clean sandbox and try running?
which version of python are you trying with?

dragon512:
@bdbaddog I removed the change to the node execute. This now is only a test fix PR.

test/packaging/guess-package-name.py
```diff
@@ -38,10 +38,14 @@
 python = TestSCons.python
 test = TestSCons.TestSCons()
 tar = test.detect('TAR', 'tar')
+bzip = test.where_is('bzip2')
```
bdbaddog, line 41: I don't think this entire test needs to be skipped if bzip2 isn't installed, only the bit around line 93

bdbaddog:
The only difference which might make sense to patch as the others in the PR have been addressed elsewhere is your change to test/packaging/guess-package-name.py

@dragon512  - can you make the sugguested change and submit a new PR?
I'm going to close this one.

test/packaging/guess-package-name.py
mwichmann: In fact, since the Python stdlib tar module can handle .bz2 tarballs, there's no reason to skip for either tar or bzip2 not found as an external command, but that would require some recoding - different PR.

PR #3167, opened 2018-08-10
Python 3.7 removed splitunc function

dmoody256:
Python 3.7 has removed splitunc function: https://docs.python.org/3/whatsnew/3.7.html#api-and-feature-removals

This PR is to handle that and maintain backwards compatibility with other python versions

## Contributor Checklist:


bdbaddog:
Feel free to check the "I have updated the appropriate documentation".
Since none needs to be updated you've completed that item as well. :)

PR #3168, opened 2018-08-11
Reduce travis texlive install footprint

dmoody256:
The travis install script install all the langs and docs for latex tools. not installing this saves hundreds of MB of download and installing, and is not used any tests anyways.

## Contributor Checklist:


bdbaddog:
So it's now only installing the bits used by the tests?

dmoody256:
yeah, should save off like a minute or two from install time during the test

bdbaddog:
O.k Thanks! I was wondering about that. Did that package get a lot bigger lately, because it didn't used to take so long..

Merging.

dmoody256:
Idk but looking back at the older builds, it may have taken off closer to 5 mins of dependency install time, with total dependency install time taking usually around 10 mins

PR #3169, opened 2018-08-12
Msvs test doesnt check for cl

dmoody256:
Sometime it is possible that Visual Studio is installed for certain version, in this case VS 2010, but different language tools may not be installed, in this case the MSVC C++ compiler. 

This happens in the Appveyor VM environment where MSVS 2010 is installed, but the C++ compiler is not installed, and fails for the test/MSVS/vs-10.0-exec.py.

This may be considered a bug in the MSVS logic, maybe it should error out if the C++ compiler is not available?

If this is not considered a bug, this change could be applied to the other msvs exec tests.

## Contributor Checklist:


bdbaddog:
Should the other *-exec.py tests be updated similarly?

dmoody256:
I sent a mail to the dev list to see if this is something we should just fix it. My PR is really just a band-aid, but yeah we could put the band-aid on the other tests as well.
https://pairlist2.pair.net/pipermail/scons-dev/2018-August/004728.html

bdbaddog:
I'll merge this one to move appveyor forward. But see email response to your email WRT should SCons also detect this case.

PR #3170, opened 2018-08-12
updated test to reflect the removal of find function from mingw tool

dmoody256:
In PR https://github.com/SCons/scons/pull/3052 the find function was removed from the mingw tool, but a few tests still used it to detect the tool. This PR uses the new generic find function to see if 'gcc' binary is on the windows installation in the default mingw paths.

## Contributor Checklist:


bdbaddog:
Good catch! Mergine.

PR #3175, opened 2018-08-13
Use travis preinstalled clang

dmoody256:
There have been issues with installing clang with apt-get and also the installation takes extra time.

Travis offers a pre-installed clang with its VM images, so updating the travis install script to use that instead.

## Contributor Checklist:


dmoody256:
Not really, I think if you build clang from source with the default install prefix, that may be where it gets installed to, but pretty much anything you build from source gets installed there, so I wouldn't say its a great default (especially for developer types, who build and install things then forget whats sitting in there usr/local/bin).

This PR is just for travis so we don't have to download and install clang.

PR #3221, opened 2018-10-14
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:

Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Also consistently use ISO 8601 date format
to be understood everywhere.
Also use gmtime to be independent of timezone.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -162,6 +162,11 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
       Three uses of variables not defined are changed.
     - Some script changes in trying to find scons engine
 
+  From Bernhard M. Wiedemann:
+    - Allow to override build date with SOURCE_DATE_EPOCH for SCons itself,
+      but not for software built with SCons.
+    - Datestamps in docs and embedded in code use ISO 8601 format and UTC
+
   From Hao Wu
     - typo in customized decider example in user guide
 
--- a/SConstruct
+++ b/SConstruct
@@ -73,7 +73,7 @@ zip = whereis('zip')
 #
 date = ARGUMENTS.get('DATE')
 if not date:
-    date = time.strftime("%Y/%m/%d %H:%M:%S", time.localtime(time.time()))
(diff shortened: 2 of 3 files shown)


bdbaddog:
Please update CHANGES.txt

So if SOURCE_DATE_EPOCH is not set, this is still changing the default format of the time/datestamp in the distibuted code right?

bdbaddog:
Also, am I understanding the motivation correctly.
That since binary distributions are produced that setting SOURCE_DATE_EPOCH you can eliminate any datestamps from being changes and so a simple (binary?) diff of packages could ensure that the package hasn't been altered

bmwiedemann:
The results should be bit-identical, so you can just compare the hash/digest of the resulting packages. We are nearly there (e.g. it is working for 95% of 11000 openSUSE packages already)

Yes, it would change the date format. If there are many parsers around, we could leave the SConstruct version as is, but I guess, few people care about the man-page header.

bdbaddog:
o.k. I'm going to leave this up for a few days and float the PR to our users mailing list.
I can't think of any good reason not to merge this though, but I like to give our users a heads up.
Maybe add in the CHANGES.txt that the effect will be datestamps in docs and embedded in code will change format even when SOURCE_DATE_EPOCH is not specified?

bmwiedemann:
The man-page format is also particularly hard to understand (without looking at the source) - e.g. is 01/02/2018 January or February?

Will add a `CHANGES.txt` entry.

bdbaddog:
US date formatting so January.

bdbaddog:
Indeed. One of many locale based date/time/etc variations.

src/CHANGES.txt
```diff
@@ -162,6 +162,10 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
       Three uses of variables not defined are changed.
     - Some script changes in trying to find scons engine
 
+  From Bernhard M. Wiedemann:
+    - Allow to override build date with SOURCE_DATE_EPOCH
```
bdbaddog: Please add a note that this is only about building SCons packages and not universal to builds done by scons when building other software.

PR #3237, opened 2018-11-07
Fix CPP PreProcess error that regex match "if" in "ifndef"

ArdaFu:
Fix the AttributeError: 'NoneType' object has no attribute 'groups' when using SCons with Python 3.7.1. 

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -104,6 +104,14 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
     - In the testing framework, module TestCommon, fixed must_contain(),
       must_not_contain(), and related methods of TestCommon class to work with
       substrings located at zero offset.
+    - Added virtualenv support. A new function Virtualenv() determines whether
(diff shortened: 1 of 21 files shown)


bdbaddog:
Can you create a test to exercise the  bug?

ArdaFu:
After tracing the code. I found that the bug is cpp regex match "if" in "ifndef".

ArdaFu:
@bdbaddog New test case is added.

src/CHANGES.txt
```diff
@@ -177,7 +177,8 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
     - Typo in customized decider example in user guide
     - Replace usage of unittest.TestSuite with unittest.main() (fix #3113)
 
-
+  From Arda Fu
+    - Fix cpp regex error that match 'if' in 'ifndef'.
```
bdbaddog: Change to be "Fix cpp scanner regex logic to treat ifndef. Previously it was not properly differentiating between if, ifdef, and ifndef"
Assuming this is accurate and I'm understanding this change..

PR #3239, opened 2018-11-13
s/get_bool_envvar/get_os_env_bool/

ptomulik:
This introduces ``get_env_bool()``, which allows to retrieve boolean values of environment variables from any dictionary-like objects. I've also renamed ``get_bool_envvar()`` to ``get_os_env_bool()``.

### DIFF
--- a/src/engine/SCons/Platform/virtualenv.py
+++ b/src/engine/SCons/Platform/virtualenv.py
@@ -37,11 +37,11 @@ virtualenv_enabled_by_default = False
 
 
 def _enable_virtualenv_default():
-    return SCons.Util.get_bool_envvar('SCONS_ENABLE_VIRTUALENV', virtualenv_enabled_by_default)
+    return SCons.Util.get_os_env_bool('SCONS_ENABLE_VIRTUALENV', virtualenv_enabled_by_default)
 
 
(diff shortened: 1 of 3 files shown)


bdbaddog:
Can you add some docstrings to the two methods?
That way when we generate API docs they'll be represented

ptomulik:
Hmm... I believe docstrings are already there. Do I miss something?

bdbaddog:
I was thinking more along the  lines of adding something like this (from PEP257)

Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)

I've been adding such as I touch functions/methods along the way. 
To improve the generated API Docs..

ptomulik:
Skimming though the file I see at least three different docstring conventions. One looks like sphinx docs, the other with ``:Parametes:``, and third is PEP257-like. I afraid, if I choose the last one, all the arguments will be put into a single line, like in [MD5collect](https://scons.org/doc/latest/HTML/scons-api/SCons.Util-module.html#MD5collect), for example. What would you suggest?

bdbaddog:
Indeed. We need to pick a standard and go with that, and then enforce it.

PR #3240, opened 2018-11-13
Fix cpp scanner regex logic to treat ifndef. This is a PY3.5+ only issue.

ArdaFu:
Previously it was not properly differentiating between if, ifdef, and ifndef

[following last PR 3237](https://github.com/SCons/scons/pull/3237)

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -186,8 +186,8 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
     - Replace usage of unittest.TestSuite with unittest.main() (fix #3113)
 
   From Arda Fu
-    - Fix cpp scanner regex logic to treat ifndef. Previously it was not properly 
-      differentiating between if, ifdef, and ifndef
+    - Fix cpp scanner regex logic to treat ifndef for py3.5+. Previously it was 
+      not properly differentiating between if, ifdef, and ifndef.
 
 
 RELEASE 3.0.1 - Mon, 12 Nov 2017 15:31:33 -0700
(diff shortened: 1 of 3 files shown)


bdbaddog:
Why did you close the last PR and make a new one? now it's hard to follow the discussion...

bdbaddog:
Your changes to the test  must not cover the fix.
If I copy it to plain master branch. it passes.

bdbaddog:
Ahh got it. This is a py3 only issue.

src/CHANGES.txt
```diff
@@ -185,6 +185,9 @@ RELEASE 3.1.0.alpha.yyyymmdd - NEW DATE WILL BE INSERTED HERE
     - Typo in customized decider example in user guide
     - Replace usage of unittest.TestSuite with unittest.main() (fix #3113)
 
+  From Arda Fu
+    - Fix cpp scanner regex logic to treat ifndef. Previously it was not properly
```
bdbaddog: Can you add that this issue only shows with python 3.5+?

ArdaFu: You mean like this?

> - Fix cpp scanner regex logic to treat ifndef for py3.5+. Previously 
      it was not properly differentiating between if, ifdef, and ifndef.

ArdaFu:
> Why did you close the last PR and make a new one? now it's hard to follow the discussion...

Sorry but I failed to rebase the official\master. Then, I reset my local repo, modified the code and force pushed to remote end. This removed my last push request commits, and closed the PR automatically. ...

bdbaddog:
Weird. Never had an issue with: git fetch upstream; git rebase upstream/master; get push -f

Seems like regex in python must've changed between py2.7 and 3.5+ to cause this issue?

ArdaFu:
Maybe, but for regex, It looks like an embedded script. Parse regex string should be same for py2.7 and py3.5+.

PR #3256, opened 2018-12-23
fix missing code contents that should be included in signature of a f…

ztessler:
…unction action

The code of a python function that is used as a build action is inspected and included in the sconsign signature so the target can be rebuilt if the function changes. However, scons currently does not notice if a constant value in the expression of a list comprehension is changed. This PR fixes that.

In this example:
```
def test(source, target, env):
    xs = [2*x for x in range(5)]
    with open(str(target[0])) as out:
        out.write(str(xs))

env = Environment()
env.Command(
    source=None,
    target='out.txt',
    action=test)
```
changing the `2` in the list comprehension in `test()` would (incorrectly) not cause a rebuild.

The existing code that computes contents of a function (in `_code_contents()`) strips off the first item from the `co_consts` tuple. In a function, this first item is the docstring (or `None` if there is no docstring). List comprehension code is not included directly in a function's `co_code`, but appears as an object in the `co_consts` tuple, and is processed on a later recursive invocation of `_code_contents()`.  List comprehensions also have their own `co_consts` tuple, though theirs do not have a docstring (or a `None` placeholder). Removing the first object from their `co_consts` effectively ignores a real constant from the code. This PR changes the logic to only remove values from the `co_consts` tuple that equal the docstring, if any.

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/Action.py
+++ b/src/engine/SCons/Action.py
@@ -258,8 +258,7 @@ def _code_contents(code, docstring=None):
     # function. Note that we have to call _object_contents on each
     # constants because the code object of nested functions can
     # show-up among the constants.
-
-    z = [_object_contents(cc) for cc in code.co_consts[1:]]
+    z = [_object_contents(cc) for cc in code.co_consts if cc != docstring]
     contents.extend(b',(')
     contents.extend(bytearray(',', 'utf-8').join(z))
     contents.extend(b')')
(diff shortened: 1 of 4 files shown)


src/engine/SCons/Action.py
```diff
@@ -258,8 +258,7 @@ def _code_contents(code, docstring=None):
     # function. Note that we have to call _object_contents on each
     # constants because the code object of nested functions can
     # show-up among the constants.
-
-    z = [_object_contents(cc) for cc in code.co_consts[1:]]
+    z = [_object_contents(cc) for cc in code.co_consts if cc != docstring]
```
bdbaddog, line 261: Is the docstring always passed in? 
Barring that if not passed in, can we grab it in this function?

bdbaddog:
Good catch.
Looks good. I added one comment regarding docstring. Please take a look

PR #3270, opened 2019-01-14
Mingw link issue

dmoody256:
The test/Win32/mingw.py test on my local system and on Appveyor 2017 image. We run the 2015 image which the test gets skipped in.

These changes allow the test to pass.

Changes in this PR:

* Update link.py to validate targets as Nodes
* Update mingw.py to remove /nologo flag added by msvc tool
* Update appveyor to use 32 bit mingw until this is done: #3291 

## Contributor Checklist:


bdbaddog:
Where is /NOLOGO getting set in the first place?
Wouldn't it be better to not set it?

Also a unit test can check this is working.

dmoody256:
> Where is /NOLOGO getting set in the first place?
> Wouldn't it be better to not set it?

I think both maybe, I suppose it matters the order the tools are loaded. It is set from the msvc.py tool, and for the test/Win32/mingw.py test, it creates a Environment(tools=['default', 'mingw']) which I think means the msvc tool runs first, adding the flag, then the mingw tool runs, removing the flag.

Another case could be Environment(tools=['mingw', 'msvc']) which would not remove the flag. I guess the msvc tool should check what tools are loaded and make sure not to add that flag if the mingw tool was loaded? The test/Win32/mingw.py should probably be updated to test this case also.

> Also a unit test can check this is working.

Ok, I'll add a unit test so we can see it passing in the Appveyor 2015 image for this PR.

bdbaddog:
Maybe add a unit test, but file a bug which says mingw and msvc stomp on each others flags?
With a short repeater?

src/engine/SCons/Tool/mingw.py
```diff
@@ -167,7 +167,8 @@ def generate(env):
 
     env['SHOBJSUFFIX'] = '.o'
     env['STATIC_AND_SHARED_OBJECTS_ARE_THE_SAME'] = 1
-
+    if 'CCFLAGS' in env:
+        env['CCFLAGS'] = SCons.Util.CLVar(str(env['CCFLAGS']).replace('/nologo', ''))
```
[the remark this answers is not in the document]
dmoody256: yeah, I was assuming it was CLVar, but its user settable so really should check if its CLVar and if not make it one first before doing the replace.

src/engine/SCons/Tool/mingw.py
```diff
@@ -150,6 +150,7 @@ def generate(env):
 
     #... but a few things differ:
     env['CC'] = 'gcc'
+    env['CCFLAGS'] = SCons.Util.CLVar('')
```
bdbaddog: with something like

`env=Environment(tools=['mslink','mingw','g++'],CCFLAGS='-myflag') `

with the new code CCFLAGS will be ignored?

bdbaddog:
Seems like the best fix here is probably to fix the tests, and then for mingw to remove any /options ? (or to issue warnings about / options)

dmoody256:
I was looking at the MSVC tool and it seemed to do the same thing, so I
thought that was ok.

On Jan 27, 2019 4:02 PM, "William Deegan" <notifications@github.com> wrote:

Seems like the best fix here is probably to fix the tests, and then for
mingw to remove any /options ? (or to issue warnings about / options)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/SCons/scons/pull/3270#issuecomment-457958409>, or mute
the thread

dmoody256:
> Seems like the best fix here is probably to fix the tests,

Not sure what you mean by fix the tests, they are explicitly testing that msvc and mingw work together:
https://github.com/SCons/scons/blob/e44c771cb3c0c5bd82ef3cd99d56cb47b2c4dc2c/test/Win32/mingw.py#L165-L166

I can add unit test such that fakes out the need for mingw to actually be present, and check that certain things were setup in the environment, like flags that don't have / in them.

> and then for mingw to remove any /options ? (or to issue warnings about / options)

It would have to remove it, because the msvc tool is adding it, and seems like the test indicates those tools should be able to be in the same environment.

src/engine/SCons/Tool/link.py
```diff
@@ -81,6 +81,8 @@ def _lib_emitter(target, source, env, **kw):
     if Verbose:
         print("_lib_emitter: target[0]={!r}".format(target[0].get_path()))
     for tgt in target:
+        if SCons.Util.is_String(tgt):
```
bdbaddog, line 84: Curious why this started failing now?

dmoody256: I don't know either, I don't know the last time we had the test/Win32/mingw.py test passing to see what changed since then.

mwichmann: A bit of a weakness of our tests which just skip a test if the tool/setup is not there is if no CI environment is set up for a particular config, it just doesn't get tested. We have potential exposure on mingw, D, docbook, TeX, SWIG, packaging.  Some of these have gotten added in recently, others fail to be added (e.g. the WiX toolset for msi packaging: apparently that code has rotted, or WiX has changed out from underneath it).

bdbaddog: Indeed.  Some tests will create fake tools to run with always and/or if the tool doesn't exist in the environment.  However in some cases that's insufficient to test if the scons code works.  Ideally we'd get a list of all tools missing and/or causing tests to be skipped at the end of a test run.. but that could be considerable work to make happen.

Any thoughts on how to better address?

bdbaddog: Yup. dev list.. :)

dmoody256:
So now with the AppendENVPath (https://github.com/SCons/scons/pull/3281) fixed, mingw.py test is having an issue because its using the 32 bit MinGW which is first in the tools default paths. Previously it would get pushed to the back and the 64 bit mingw would be next in line. Appveyor sets an environment path for 64 bit mingw, which then fails to run the 32 bit executable. It seems we would need to update the appveyor path to point to 32 bit MinGW.

Which begs the question, what if someone had multiple versions of a tool installed, how would they force scons to make sure the desired one is chosen? Seems like they would need to call PrependENVPath with which ever one they wanted after the tool was loaded.

mwichmann:
> Which begs the question, what if someone had multiple versions of a tool installed, how would they force scons to make sure the desired one is chosen? Seems like they would need to call PrependENVPath with which ever one they wanted after the tool was loaded.

I've wondered that... having several java versions installed isn't that uncommon (although maybe multiple jdks is less common).  We include JAVA_HOME as an argument to the build to help with that.

bdbaddog:
This is what we added TARGET_ARCH for (well 64 vs 32 bits), but currently
it's only used for MSVC.
Is there an easy way to determine if the installled mingw is 32 vs 64 bits
and could be used to select the correct installed version?
Also MINGW_VERSION maybe to say we only want a specific version (also
copying from MSVC)?

-Bill

src/engine/SCons/Tool/link.py
bdbaddog: Curious though this issue should've affected all linkers and not just mingw... if they all needed to be turned into nodes at this point?  Does mingw fail if you don't do this bit?

dmoody256: Yeah, it will get an exception that 'str' has no member 'attributes' just below. I'm not sure if there is somewhere further up the chain that the target should have been converted.

dmoody256: Still should be some kind of check here even if the real fix is somewhere else. Maybe a try/except? Add a warning message? I think assuming its a file if its a string is a pretty good default case.

bdbaddog: Yes. Can you change to try/except and put a comment in the except that we'd expect the target to be a Node at this point.  And a # TODO to track down the why at some point?
I don't want to lose this discovery without some reasonable way of remembering to address it at some point if possible.

dmoody256: I chased this down and found that mingw is getting an emitter list of [mslink.py winLibEmitter, link.py shlib_emitter, mingw.py shlib_emitter]. mslink's winLibEmitter never converts its new targets to File's and just leaves them as strings:
https://github.com/SCons/scons/blob/9f090132091e56c123262106f3563142787c2fcf/src/engine/SCons/Tool/mslink.py#L144-L154
The other tools with emitters take care to convert targets to File objects:
https://github.com/SCons/scons/blob/9f090132091e56c123262106f3563142787c2fcf/src/engine/SCons/Tool/cyglink.py#L100-L106
So the fix can be applied by converting the the targets in mslink.py's winLibEmitter to File's. Probably still want to have a check in link.py to convert str's as well.

dmoody256: I left the if check instead of a try except because the current code looks cleaner than:
```python
 for tgt in target:
        try:
            tgt.attributes.shared = 1
        except AttributeError:
            tgt = env.File(tgt)
            tgt.attributes.shared = 1
```
I am not sure about the speed up. Let me know your opinion.

bdbaddog:
Merging.

PR #3280, opened 2019-01-27
tools should not override existing path order

dmoody256:
This is a potential solution to https://github.com/SCons/scons/issues/3276.

This sets all tools to not delete existing paths, maintaining the original environment.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -11,6 +11,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
     - Whatever John Doe did.
 
+  From Daniel Moody:
+    - Update tools to not delete existing paths in the PATH when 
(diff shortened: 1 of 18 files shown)


bdbaddog:
Is there anywhere left in scons where delete_existing is defaulted?
Perhaps it makes more sense to change the default?
I can't imagine that the "normal" usage would want to move it to the end of a given path?

dmoody256:
> Is there anywhere left in scons where delete_existing is defaulted?
> Perhaps it makes more sense to change the default?
> I can't imagine that the "normal" usage would want to move it to the end of a given path?

I made another to test that out PR #3281

dmoody256:
I think I will close this PR in favor of changing the default and updating the test in #3281?

bdbaddog:
O.k. Closing.

PR #3281, opened 2019-01-28
Change appendenvpath default

dmoody256:
Potential solution for https://github.com/SCons/scons/issues/3276

Change the default to not touch the path.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -11,6 +11,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
     - Whatever John Doe did.
 
+  From Daniel Moody:
+    - Change the default for AppendENVPath to delete_existing=0, so path
+      order will not be changed, unless explicitly set (Issue #3276)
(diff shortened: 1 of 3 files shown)


mwichmann:
Looks like there's a test that depends on the old default behavior.

dmoody256:
I checked the related documentation in the man page, and it looks like its correct already, so that means the code before this PR was not in line with the documentation.

bdbaddog:
I thought the docs didn't specify the default for delete_.. parameter?

dmoody256:
> I thought the docs didn't specify the default for delete_.. parameter?

They don't say an explicit default value, but they do describe the default behavior as:
> _This will only add any particular path once (leaving the last one it encounters and ignoring the rest, to preserve path order)_

bdbaddog:
So this is ready to merge right?
mingw failure expected at this point?

dmoody256:
Yeah, this ones good to go

PR #3326, opened 2019-03-07
Add /nologo flag to RCFLAGS

maciek-codes:
Add /nologo parameter to RCFLAGS to avoid the banner during resource compilation.

## Contributor Checklist:


bdbaddog:
Is this a preference or does the rc compiler fail if you don't specify?

maciek-codes:
The RC compiler doesn't fail. It just shows a banner each time it's invoked
and spams the output. We set /nologo by default in the CCFLAGS so I assumed
this was just missed and should be a default too.

bdbaddog:
That seems reasonable as we have /nologo for almost all the other MS build tools.
Please add a blurb to src/CHANGES.txt

Then we should be able to merge.

bdbaddog:
Merged. Thanks!

Issue #2811, opened 2018-01-02
Dependency order problem, rebuilds due to incorrect implicit dependencies


bdbaddog:
This issue was originally created at: 2012-01-10 10:19:23.
This issue was reported by: `dirkbaechle`.

> dirkbaechle said at 2012-01-10 10:19:23

Hi there,
 
as discussed at some length in http://scons.tigris.org/ds/viewMessage.do?dsForumId=1268&dsMessageId=2855320, SCons starts rebuilds on a second call, when obviously nothing has changed.
> 
>I started to debug this problem, with gem5 ( www.gem5.org ) as example, and it is connected to the way we scan (don't scan) for implicit dependencies. I am able to reproduce the "2nd build" on my Ubuntu 11.04 box, by repeatedly calling:
```console
rm -f build/sconsign.dblite
scons build/ALPHA_SE/python/m5/internal/param_NetworkTest.i
scons build/ALPHA_SE/python/m5/internal/param_NetworkTest.i
```

Note that on an initial build, or after the "rm -f sconsign.dblite", no signature infos are available at all.
 
So, when the Taskmaster tries to find the next "ready" node it uses `get_implicit_deps()` to find all children for the node in question. But no children are returned, because `current_check()` (`SCons/Scanner/__init__.py:318`) checks whether the node is "up_to_date". No signature info means that the node has changed, and then `__call__()` (`SCons/Scanner/__init__.py:210`) returns the emtpy list:
```py
        if self.scan_check and not self.scan_check(node, env):
            return []
```
 
Stack trace for this is:
 
```
MainThread - pid30609_seq1
    changed [__init__.py:1066]
    is_up_to_date [FS.py:3028]
    current_check [__init__.py:318]
    __call__ [__init__.py:210]
    get_found_includes [FS.py:2681]
    get_implicit_deps [__init__.py:555]
    scan [Executor.py:482]
    scan_sources [Executor.py:451]
    scan [__init__.py:648]
    children [__init__.py:999]
    get_all_children [Executor.py:261]
    _find_next_ready_node [Taskmaster.py:786]
    next_task [Taskmaster.py:918]
    start [Job.py:193]
    run [Job.py:111]
    _build_targets [Main.py:1264]
    _main [Main.py:1070]
    _exec_main [Main.py:1306]
    main [Main.py:1342]
    <module> [Main.py:1402]
    run [pydevd.py:916]
    <module> [pydevd.py:1145] 
```
[...]

PR #3340, opened 2019-04-02
Fix issue #2811 spurious rebuilds due to incorrect waiting_parents on nodes when builder has more than one target and source file generated

bdbaddog:
Fix Issue #2811 

Reproducing this issue requires the following.
1. Generated source file which outputs 2 (or more) files
2. Action string gets scanned providing only compiler as part of implicit scan
3. Generated file gets built. Without the bugfix only the first target's .implicit list is cleared.
4. builder/executor/action gets tried again and implicits scanned. 2nd to Nth targets end up 
   with the compiler at the beginning of the implicit list and the rest of the scanned files added to that list.
5. That bimplicit gets saved into sconsign
6. Second run loads sconsign, now with generated file present a regular implicit scan occurs. This yields 2nd through
   Nth target's implicit lists changing when compared to SConsign's which have been loaded.
7. This forces rebuild of source file and this propagates to massive recompile
## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/Node/__init__.py
+++ b/src/engine/SCons/Node/__init__.py
@@ -103,9 +103,9 @@ implicit_deps_changed = 0
 
 # A variable that can be set to an interface-specific function be called
 # to annotate a Node with information about its creation.
-def do_nothing(node): pass
+def do_nothing_node(node): pass
 
-Annotate = do_nothing
+Annotate = do_nothing_node
 
 # Gets set to 'True' if we're running in interactive mode. Is
 # currently used to release parts of a target's info during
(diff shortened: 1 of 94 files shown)


src/engine/SCons/Node/__init__.py
```diff
@@ -784,6 +785,25 @@ def built(self):
         for parent in self.waiting_parents:
             parent.implicit = None
 
+            # Handle issue where builder emits more than one target and
```
garyo, line 788: This looks fine to me. I guess my only question would be why aren't the other targets already in waiting_parents so they get cleared above? (Just curious)

bdbaddog: It's a combination of the first target hit by the taskmaster walk and the fact that the N targets aren't linked.  I'm going to continue looking at this, but this fix resolves a long standing issue so accepting the good enough, and then will continue working on the maybe better solution.

PR #3376, opened 2019-05-18
pytar, sdist tool from old hg repository

dholth:
This is the Python stdlib tar implementation from my 2016 mercurial pull request, plus a src_sdist packging tool.

As of PEP517 sdists are supposed to be PAX-format tar, additionally it's nice to omit the account name and uid from these files.

I might bundle this in enscons instead, but there isn't a way to add formats to SCons.Tool.packaging without monkey patching, is there?

### DIFF
--- a/src/engine/SCons/Tool/pytar.py
+++ b/src/engine/SCons/Tool/pytar.py
@@ -1,0 +1,116 @@
+"""SCons.Tool.pytar
+
+Tool-specific initialization for tar (Python stdlib implementation)
+
+There normally shouldn't be any need to import this module directly.
+It will usually be imported through the generic SCons.Tool.Tool()
+selection method.
+
+"""
+
+#
+# __COPYRIGHT__
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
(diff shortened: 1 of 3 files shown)


src/engine/SCons/Tool/pytar.py
```diff
@@ -0,0 +1,116 @@
...
+    targid = env.get('TARGID')
+    tarmtime = env.get('TARMTIME')
+    with tarfile.TarFile.open(
+            str(target[0]),
+            'w' + (compression and ':') + compression,
+            format=tarformat) as tar:
+
+        def _filter(info):
+            """Return potentially anonymize tarinfo"""
+            if taruid != None:
```
mwichmann, line 61: The style checker that runs as part of the CI system objects to the form of the comparison, that's why it has marked this PR as a failed build.

Flake8
[E711] comparison to None should be 'if cond is not None:'

mwichmann:
The Appveyor build fail is one of the transients we see from time to time, don't worry about it.

I'd put a small amount of money on @bdbaddog asking for a test to go with this...

bdbaddog:
I've been thinking about this patch.
On windows due to race condition issues we've seen builds die when python
should have closed a file but the consumer finds the file still opened by a
process. (I believe this is due to python releasing the GIL around file
closing).

Anyway, it might be better to have a python script which scons will run if
there is no tar exe found instead. This way no issue with file handles
still open in process.

Thoughts?

dholth:
The with statement doesn't work?

I think pure python in process tar is right for scons, but what I'd like for me is a way to customize the packaging tool, at a minimum add formats without monkey patching, and change its build directory; these would solve the enscons sdist issues.

bdbaddog:
shutil.copy() causes the issue. Swapping in win32 CopyFile resolves the
issue.
So I'm pretty sure the issue is multi-thread + python's c code for close
releasing the GIL.
I've not yet created a repro to submit to python devels.

Bring the other packaging issues to the dev mailing list?

dholth:
This was my solution. With a tar based on the standard library I can prepend the dubiously named `$TARPREFIX` (put everything in a subdirectory) to each file without actually copying all the files into a subdirectory, and can stop using the `Packaging` tool entirely. https://bitbucket.org/dholth/enscons/commits/132270824a7ec4f4d6021ef0c1286944044b90ef

PR #3377, opened 2019-05-18
Use TempFileMunge to link and compile under posix systems

KastB:
Scons doesn't use TempFileMunge to link targets under posix systems at the moment. This results in "argument list too long" errors for larger projets.
This commit solves this bug/adds the feature to handle larger projets

## Contributor Checklist:

* [x] I have created a new test or updated the unit tests to cover the new/changed functionality. (already covered)
* [x] I have updated the appropriate documentation (no additional documentation required => rather a bug fix than a new feature)

### DIFF
--- a/src/engine/SCons/Tool/link.py
+++ b/src/engine/SCons/Tool/link.py
@@ -312,7 +312,7 @@ def generate(env):
 
     env['SHLINK'] = '$LINK'
     env['SHLINKFLAGS'] = SCons.Util.CLVar('$LINKFLAGS -shared')
-    env['SHLINKCOM'] = "${TEMPFILE('$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'), '$SHLINK')}"
+    env['SHLINKCOM'] = "${TEMPFILE('$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS', '$SHLINK')}"
 
     # don't set up the emitter, because AppendUnique will generate a list
     # starting with None :-(
--- a/src/engine/SCons/Tool/link.py
+++ b/src/engine/SCons/Tool/link.py
@@ -312,7 +312,7 @@ def generate(env):
 
     env['SHLINK'] = '$LINK'
     env['SHLINKFLAGS'] = SCons.Util.CLVar('$LINKFLAGS -shared')
-    env['SHLINKCOM'] = '$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'
+    env['SHLINKCOM'] = "${TEMPFILE('$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'), '$SHLINK')}"
 
     # don't set up the emitter, because AppendUnique will generate a list
     # starting with None :-(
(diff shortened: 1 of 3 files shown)


mwichmann:
You've left unaltered the placeholder in the PR which says to describe the change, which Bill will ask you to update ;)

CI builds all failed, errors looking like this:
```
scons: *** [libbar.so] SyntaxError `unexpected EOF while parsing (<string>, line 1)' trying to evaluate `${TEMPFILE('$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'), '$SHLINK')}'
```

KastB:
It would be great if someone could explain why a test case on windows fails for changes that should only affect linking on posix systems.
Actually the outputs look pretty similar to me - only the suffix changed from cpp to obj.
Is this some kind of bad luck or did I really bring in a bug?
Can someone give me a hint about what I did wrong?

KastB:
This test regularly fails with the same error message (cpp instead of obj)
https://ci.appveyor.com/project/SCons/scons/builds/24643086/job/6o5koguwgc1cof7t
https://ci.appveyor.com/project/SCons/scons/builds/24669659/job/iwu6ntxc06wrsjaw
https://ci.appveyor.com/project/SCons/scons/builds/24514051/job/incju08ick90059n

Is this a faulty test, that only works by chance or is it a test with high sensitivity that detects a lot of different bugs?

mwichmann:
That one's not your problem... that test has been intermittently barfing since we added more parallelism to the CI runs, and I believe the "fail" is that lines are coming out in a slightly different order. than the test exepcts.

KastB:
What can I do to make the green hook appear?
Can I re-trigger a test run?
Or is this failure just ignored?
Its kind of frustrating - I think even my second version of the pull request would have worked but failed at that test. Since then I did reverts and desperate changes in the hope it will work again.

Probably I'm not the one to change that test.
Despite there would probably a pragmatic solution to ignore the suffix of the file in the checked output:
something like this:
 search re='^scons>>> .*foo\.cpp.*$'
=> search re='^scons>>> .*foo\..*$'

KastB:
Ah ok great - so I'm done for the moment ;)
Thank you for your explanations.

bdbaddog:
Yes ignore the failure of the test Mats and you are talking about.

bdbaddog:
I'm not sure this is safe to use on all tools which inherit from SCons.Tool.link, which linkers are you trying to resolve this for?

Most likely you should be doing this on a tool by tool basis.
For example, if you're using gnulink (most likely).. then make this change in src/engine/SCons/Tool/gnulink.py

src/engine/SCons/Tool/link.py
```diff
@@ -312,7 +312,7 @@ def generate(env):
 
     env['SHLINK'] = '$LINK'
     env['SHLINKFLAGS'] = SCons.Util.CLVar('$LINKFLAGS -shared')
-    env['SHLINKCOM'] = '$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS'
+    env['SHLINKCOM'] = "${TEMPFILE('$SHLINK -o $TARGET $SHLINKFLAGS $__SHLIBVERSIONFLAGS $__RPATH $SOURCES $_LIBDIRFLAGS $_LIBFLAGS', '$SHLINK')}"
```
bdbaddog, line 315: should the second argument be SHLINKCOMSTR ?

bdbaddog: ?any updates

bdbaddog:
No updates on this in a while. We'd need to have this only set per compiler as not all compilers can take a command file on command line.  I'm going to close this. Please reopen if you get around to making this change.

acmorrow:
@KastB - If you want to see an example of how to set this up in your project without requiring SCons to do so itself, the MongoDB SConstruct does this: https://github.com/mongodb/mongo/blob/e609eacb52a443d8d0332b2417184e67499a7da7/SConstruct#L1598-L1616

PR #3386, opened 2019-06-05
[WIP] Try to keep the number of jobs as high as possible

acmorrow:
## Remove this paragraph

## Contributor Checklist:

### DIFF
--- a/SCons/Job.py
+++ b/SCons/Job.py
@@ -31,6 +31,7 @@ import SCons.compat
 
 import os
 import signal
+import sys
 
 import SCons.Errors
 import SCons.Warnings
@@ -306,9 +307,9 @@ else:
             """Put task into request queue."""
             self.requestQueue.put(task)
 
-        def get(self):
+        def get(self, block=True):
             """Remove and return a result tuple from the results queue."""
-            return self.resultsQueue.get()
+            return self.resultsQueue.get(block=block)
 
         def preparation_failed(self, task):
             self.resultsQueue.put((task, False))
@@ -379,11 +380,12 @@ else:
             jobs = 0
 
             while True:
-                # Start up as many available tasks as we're
-                # allowed to.
-                while jobs < self.maxjobs:
+
+                # Process jobs until we find one that requires
+                # execution, and enqueue it.
+                while True:
                     task = self.taskmaster.next_task()
-                    if task is None:
+                    if not task:
                         break
 
                     try:
@@ -398,17 +400,41 @@ else:
                             # dispatch task
                             self.tp.put(task)
                             jobs += 1
+                            break
                         else:
                             task.executed()
                             task.postprocess()
 
-                if not task and not jobs: break
+                # If 'task' is true-ish, then we enqueued a new task,
+                # otherwise we didn't. If we didn't enqueue a task and
+                # we don't have any jobs pending, then there is
+                # nothing left to do.
+                if not (task or jobs):
+                    break
 
-                # Let any/all completed tasks finish up before we go
-                # back and put the next batch of tasks on the queue.
-                while True:
-                    task, ok = self.tp.get()
-                    jobs -= 1
+                # If we couldn't find any new tasks requiring
+                # execution, we need to wait for a completion.
+                block = (task is None)
+
+                # If we have reached our quota for outstanding jobs,
+                # we must wait for at least one job to complete before
+                # re-entering the loop above.
+                block |= (jobs == self.maxjobs)
+
+                while jobs:
+                    try:
+                        task, ok = self.tp.get(block=block)
+                        jobs -= 1
+                    except queue.Empty:
+                        # We did a non-blocking wait, but there was
+                        # nothing ready yet. See if we can find another
+                        # task to enqueue. If we can't we will end up
+                        # back in this loop but we will do a blocking
+                        # wait.
+                        break
+
+                    # We don't want to block more than once.
+                    block = False
 
                     if ok:
                         task.executed()
@@ -425,9 +451,6 @@ else:
                         task.failed()
 
                     task.postprocess()
-
-                    if self.tp.resultsQueue.empty():
-                        break
 
             self.tp.cleanup()
             self.taskmaster.cleanup()
--- a/SCons/Job.py
+++ b/SCons/Job.py
@@ -366,7 +366,7 @@ else:
 
             self.taskmaster = taskmaster
             self.interrupted = InterruptState()
-            self.tp = ThreadPool(num, stack_size, self.interrupted)
+            self.thread_pool = ThreadPool(num, stack_size, self.interrupted)
 
             self.maxjobs = num
 
@@ -406,7 +406,7 @@ else:
                     else:
                         if task.needs_execute():
                             # dispatch task
-                            self.tp.put(task)
+                            self.thread_pool.put(task)
                             jobs += 1
                             retired -= 1
                             if retired == 0:
@@ -434,7 +434,7 @@ else:
                 retired = 0
                 while jobs:
                     try:
-                        task, ok = self.tp.get(block=block)
+                        task, ok = self.thread_pool.get(block=block)
                         jobs -= 1
                         retired += 1
                     except queue.Empty:
@@ -464,7 +464,7 @@ else:
 
                     task.postprocess()
 
-            self.tp.cleanup()
+            self.thread_pool.cleanup()
             self.taskmaster.cleanup()
 
 # Local Variables:
--- a/SCons/Job.py
+++ b/SCons/Job.py
@@ -409,8 +409,6 @@ else:
                             self.thread_pool.put(task)
                             jobs += 1
                             retired -= 1
-                            if retired == 0:
-                                break
                         else:
                             task.executed()
                             task.postprocess()
--- a/SCons/Job.py
+++ b/SCons/Job.py
@@ -377,13 +377,22 @@ else:
             more tasks. If a task fails to execute (i.e. execute() raises
             an exception), then the job will stop."""
 
+            retired = 0
             jobs = 0
 
             while True:
 
-                # Process jobs until we find one that requires
-                # execution, and enqueue it.
-                while True:
+                # Try to enqueue one more job than we retired on the
+                # prior cycle so that we can increase parallelism over
+                # time if the DAG and the throughput allow it.
+                retired += 1
+
+                # Process tasks, adding execution jobs for tasks
+                # requiring execution until we either have hit the
+                # maximum number of jobs allowed, or have started one
+                # more than as many as we retired on the prior cycle,
+                # or become unable to find another task.
+                while jobs < self.maxjobs and retired > 0:
                     task = self.taskmaster.next_task()
                     if not task:
                         break
@@ -400,7 +409,9 @@ else:
                             # dispatch task
                             self.tp.put(task)
                             jobs += 1
-                            break
+                            retired -= 1
+                            if retired == 0:
+                                break
                         else:
                             task.executed()
                             task.postprocess()
@@ -421,10 +432,12 @@ else:
                 # re-entering the loop above.
                 block |= (jobs == self.maxjobs)
 
+                retired = 0
                 while jobs:
                     try:
                         task, ok = self.tp.get(block=block)
                         jobs -= 1
+                        retired += 1
                     except queue.Empty:
                         # We did a non-blocking wait, but there was
                         # nothing ready yet. See if we can find another
--- a/SCons/Job.py
+++ b/SCons/Job.py
@@ -31,7 +31,6 @@ import SCons.compat
 
 import os
 import signal
-import sys
 
 import SCons.Errors
 import SCons.Warnings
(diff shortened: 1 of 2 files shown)


acmorrow:
@bdbaddog - Per discussions from a month or so ago, here is a WIP PR about some ideas I had for a different way to schedule jobs.

This came from a long ago discussion where I'd observed something of a sawtooth pattern of job execution. In the existing implementation of Job.py, we queue up N active jobs, and then reap all the results. As a result, the number of concurrent jobs rises towards N until the loop can't find more work, then falls back to zero as we consume all the enqueued results, then climbs back towards N, etc.

One consequence of this is that if you have a distributed compile cluster (icecream, etc.), and you build with a very large N, you can actually spend a lot of time trying to build up to N jobs, even though SCons can't actually drive that much concurrency due to the latency of finding the next task. Then, concurrency drops to zero as we reap perhaps several hundred pending jobs.

Had we actually reaped some of those jobs that we had already started, this might have been better, as it would have unblocked other jobs, potentially allowing us to better exploit the available concurrency.

The approach here works differently. It enqueues just one job and then goes to reap the results. If results are available we process all of them, otherwise, we go back to enqueueing. If we can't find any new jobs, or if we have as many jobs as we are allowed, then we do a blocking get instead, since we either must consume results to unblock additional work, or are prohibited from executing any more.

The overall result is that when the -j argument is something we can actually achieve, we avoid the sawtooth pattern because we toggle back and forth between enqueueing and dequeuing jobs. With some printf debugging, I find that the number of executing jobs is usually very close to the -j value, with out sudden drop offs towards zero.

One consequence of this model is that if you request far more jobs than you can actually execute, you may never actually have that many jobs executing. In other words, the -j value becomes aspirational. For me, building with -j1000 on a machine with 8 cores never gets past 25 jobs in flight.

mwichmann:
I was going to say: I bet you break some of the more fragile tests.  This is what Travis thinks:

```
8/1202 (0.67%) /home/travis/virtualenv/python2.7.14/bin/python -tt src/engine/SCons/JobTests.py
F......
FAIL: runTest (__main__.ParallelTestCase)
test parallel jobs
  File "src/engine/SCons/JobTests.py", line 315, in runTest
    assert ThreadPoolCallList in expect, ThreadPoolCallList
AssertionError: ['put(1)', 'put(2)', 'put(3)']
```

acmorrow:
@mwichmann - Oh for sure. I've made zero efforts to get the tests running. I actually implemented htis directly inside the vendored copy of SCons inside MongoDB and just copied it into my SCons clone and made a PR. I didn't see much value in adjusting the tests unless there is some agreement that this is interesting. Then I'll be happy to work on them.

mwichmann:
This sounds like a generally sensible approach, given that I don't actually know what's going on under the covers.  The more "normal" model for concurrent tasks is to fill up a queue, and then have workers pull things off the queue as they're ready for more work - but I gather scons doesn't work like that? The concept of this change seems to get a bit closer to that, where finished jobs are reaped closer to when they're done, clearing space for new work more quickly - which can only be a good thing?

src/engine/SCons/Job.py
```diff
@@ -375,55 +375,83 @@ def start(self):
...
+                  break
+
+
+               # If we are at maxjobs, then force a wait for a job
+               # completion so that we don't go over the limit by
+               # looping back up to next_task. If we didn't find any
+               # new tasks above, we must also block so that we don't
+               # busy loop.
+               should_block = False
+               should_block |= (jobs == self.maxjobs)
```
grossag: As I mentioned on my proposed #3404 , I think we should be more aggressive than this. I would like for us to not block until task is None. Any time blocking on the main Python thread is wasted time where we could be gathering more tasks for dispatch.

acmorrow: @grossag - This only blocks if we were either unable to find any new tasks to run, or if we are not allowed to issue any more jobs. In the former case, we won't be able to find any new tasks to run until we retire some jobs that we have issued, and processed their results unblocking access to new parts of the DAG. So I think it does what you are suggesting.

grossag: But this also blocks if there are tasks available to run but we are not able to issue any more jobs. A more efficient scheduler would actually keep calling taskmaster.next_task() if jobs==self.maxjobs and put them in a list. This is because next_task() is not instantaneous in that it generates scan requests, which goes to disk. Any time spent blocking while next_task() would return more tasks is time wasted.

grossag: Here is a scheduler I wrote as part of remote caching support. Pull request is coming eventually, but I need to improve testability. This scheduler will focus on doing three things as efficiently as possible: (1) gather tasks, (2) fetch tasks from remote cache, and (3) deploy jobs to the thread pool. It will only wait on jobs results if there are no tasks left to gather and no fetches pending to the remote cache. Here is the code: https://gist.github.com/grossag/9e7ef2e8c3c02584152c67812af7cd3a .

I post this because it demonstrates what I mean about not waiting on jobs unless taskmaster.next_task() returns None.

acmorrow:
@bdbaddog - I've updated this to fix the failing test.

src/engine/SCons/Job.py
acmorrow: @grossag - OK, yes, I see what you are saying. However, this is a meaningful improvement over the current behavior, which exhibits unfortunate sawtooth behavior. Would you be OK with seeing this merge now, even if it is later supplanted by your even better approach? Keeping things iterative, etc.

acmorrow: @grossag - Actually, having thought about this a little more, I think it is easy to apply that idea to this implementation. I'll give it a try.

grossag:
I saw your note last night about being more aggressive in collecting tasks instead of blocking. To me, that is where the real gains will happen.

I say that because I don't agree with the underlying premise that the Parallel job scheduler has a sawtooth behavior. If that was true, it would indeed be a catastrophic scheduler problem where one long job can hurt build times in a disproportionate manner. After all, it would be very bad if we had room for 16 jobs but we were sitting with only 1 active job and waiting for that last job to be done instead of enqueueing15 more jobs.

This check https://github.com/SCons/scons/blob/master/src/engine/SCons/Job.py#L425 prevents sawtooth behavior. By checking whether resultsQueue is empty, we effectively break out of the loop when there are no more job results available. That's exactly what we want. This loop says "I will process all results until we don't have any results left." That could be all N results if they are all done, but usually it will be 1 or 2 results and then we will go back to gathering more tasks.

acmorrow:
I made a plot (using GNU graph, so, sorry it isn't prettier).

plot

I created this by applying the following diff to jobs.py:

```
diff --git a/src/third_party/scons-3.1.2/scons-local-3.1.2/SCons/Job.py b/src/third_party/scons-3.1.2/scons-local-3.1.2/SCons/Job.py
index 214a0c9fb8..c5009ddbbc 100644
--- a/src/third_party/scons-3.1.2/scons-local-3.1.2/SCons/Job.py
+++ b/src/third_party/scons-3.1.2/scons-local-3.1.2/SCons/Job.py
@@ -35,6 +35,7 @@ import SCons.compat

 import os
 import signal
+import sys

 import SCons.Errors

@@ -394,6 +395,7 @@ else:
                             # dispatch task
                             self.tp.put(task)
                             jobs = jobs + 1
+                            print('XXX JOBS ', jobs, file=sys.stderr)
                         else:
                             task.executed()
                             task.postprocess()
@@ -405,6 +407,7 @@ else:
                 while True:
                     task, ok = self.tp.get()
                     jobs = jobs - 1
+                    print('XXX JOBS ', jobs, file=sys.stderr)

                     if ok:
                         task.executed()
```
I then ran our build as follows:

`/usr/bin/time /opt/mongodbtoolchain/v3/bin/python3 buildscripts/scons.py ... -j 300 install-core 2>baseline.log`

Then plotted the results with:

`graph -T PNG <(grep JOBS baseline.log  | awk '{ print NR, $3 }') > plot.png`

acmorrow:
Actually, that first graph got cut off too early due to some noise from compiler warnings that also went to stdout. Here is a complete graph of the baseline run:

plot

And here is a plot of the run with the currently proposed changes in this CR:

plot2

This one doesn't implement your observation about having the SCons main thread never block. I'll post updated graphs when I have that working to my satisfaction.

In any event @grossag, I think it is clear that we agree that the scheduler can be better than it currently is.

acmorrow:
Updated so that the SCons main thread is never blocked unless no tasks could be found and no completion events are queued. This breaks the tests again, I'll take a look at fixing them, but it may be harder this time.

src/engine/SCons/Job.py
```diff
@@ -375,15 +376,16 @@ def start(self):
...
+                # execution, and enqueue it.
+                while True:
                     task = self.taskmaster.next_task()
-                    if task is None:
+                    if not task:
                         break
 
                     try:
-                        # prepare task for execution
+                        # pepare task for execution
```
acmorrow: Not sure how this happened but I've fixed it.

bdbaddog:
@acmorrow - What's the status on this?

bdbaddog:
@acmorrow - any updates on this?

acmorrow:
@bdbaddog - I've rebased this and force pushed it so it is modern w.r.t. the tree because I'm interested in this work again. I know that @grossag has a competing approach and rewritten task scheduler, but perhaps we can look into merging this as an intermediate state until that work supplants it? It does seem to result in meaningfully better behavior for the MongoDB build. Effectively, trying so hard in the initial loop to find executable jobs up to the concurrency limit doesn't work well when the concurrency limit is very high, for things like icecream builds.

acmorrow:
I pushed another commit here, because I had an idea. One of my concerns with this was always that you might not achieve as much parallelism as you could, because it would only ever try to add one job on each pass through. You could get locked into a steady state where each cycle you added one and retired one, but you might have been able to add more.

The second commit attempts to address that concern by not just bailing on the "enqueue" phase of the loop after discovering one task that needs execution, but by attempting to enqueue up to as many jobs were retired on the prior round of job retirement, plus one for growth.

Once maybe nice feature of this is that you are somewhat protected from running with -j1000 or similar on a local build, because your local machine will never be retiring 1000 jobs at once, so you will never try to enqueue that many. Anyway, something to think about.

SCons/Job.py
```diff
@@ -396,19 +406,47 @@ def start(self):
...
                         else:
                             task.executed()
                             task.postprocess()
 
-                if not task and not jobs: break
+                # If 'task' is true-ish, then we enqueued a new task,
+                # otherwise we didn't. If we didn't enqueue a task and
+                # we don't have any jobs pending, then there is
+                # nothing left to do.
+                if not (task or jobs):
```
bdbaddog, line 422: So this is now the only way to exit the `while jobs:` loop?

SCons/Job.py
```diff
@@ -396,19 +406,47 @@ def start(self):
...
+                        task, ok = self.thread_pool.get(block=block)
+                        jobs -= 1
+                        retired += 1
+                    except queue.Empty:
+                        # We did a non-blocking wait, but there was
+                        # nothing ready yet. See if we can find another
+                        # task to enqueue. If we can't we will end up
+                        # back in this loop but we will do a blocking
+                        # wait.
+                        break
```
bdbaddog, line 444: So this is the only exit from this loop now?
(Vs before it would also check if results queue was empty).

bdbaddog: Which will only happen when block = False, which is only set if no new tasks were added and we're not already at max jobs.

(not counting users hit's control-c or similar)

acmorrow: Right, only way out is to hit `break` here, which only happens on the non-blocking case. But, if you did a blocking wait, you will come back around and do a non-blocking one, since we unconditionally set `block` to `false` if we don't `break`.
- If you are non-blocking, an empty queue takes you directly to the `break`.
- If you are blocking, your next attempt after unblocking will be non blocking, and an empty queue takes you to `break`.

acmorrow: Also, the prior check of whether the results queue was empty is basically equivalent to doing the non-blocking `get`.

SCons/Job.py
```diff
@@ -396,19 +406,47 @@ def start(self):
                     else:
                         if task.needs_execute():
                             # dispatch task
-                            self.tp.put(task)
+                            self.thread_pool.put(task)
                             jobs += 1
+                            retired -= 1
+                            if retired == 0:
```
acmorrow: I think this `if retired == 0` is redundant with the loop guard at the top, btw, so it should probably be dropped.

bdbaddog: yup. I believe your are correct!

acmorrow: Fixed!

acmorrow:
@bdbaddog - I ran some A/B tests of this in our icecream cluster last night. It is a little faster, but not dramatically so. While watching the behavior of the build cluster in the icecream monitor, it looks to me like the way `retired` works isn't quite good yet: it is too conservative in ramping up, and too aggressive in ramping back down. I'd like to flip that: more aggressive in ramping up, less aggressive in ramping down. I'm going to see if I can come up with something simple that behaves closer to what I had in mind and re-run my performance experiments.

acmorrow:
I'm closing this in favor of #4222.

PR #3403, opened 2019-07-02
LaTeX scanner: Find > 1 includes per line

luhk:
`^[^%\n]*` at the beginning of the reg ex would match all but the last include so that they were lost. There is no point checking for `%` since comments are stripped anyways (and this would also fail on lines containing an escaped percent sign `\%`, right?). Also, I don't see the point of checking for the newline, so I just removed that altogether.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -42,9 +42,15 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       which specifies what character to join all the argements output into the tempfile. The default remains a space
       when mslink, msvc, or mslib tools are loaded they change the TEMPFILEARGJOIN to be a line separator (\r\n on win32)
 
-    From Michael Hartmann:
+  From Michael Hartmann:
     - Fix handling of Visual Studio Compilers to properly reject any unknown HOST_PLATFORM or TARGET_PLATFORM
 
+  From Mathew Robinson:
+    - Update cache debug output to include cache hit rate.
+    - No longer unintentionally hide exceptions in Action.py
+
+  From Lukas Schrangl:
+    - Enable LaTeX scanner to find more than one include per line
 
   From Mats Wichmann:
     - scons-time takes more care closing files and uses safer mkdtemp to avoid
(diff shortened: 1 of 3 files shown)


bdbaddog:
Please update src/CHANGES.txt to add a blurb for this change.

luhk:
OK, done that (and also restored alphabetical order).

src/CHANGES.txt
```diff
@@ -42,9 +42,15 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
...
-    From Michael Hartmann:
+  From Michael Hartmann:
     - Fix handling of Visual Studio Compilers to properly reject any unknown HOST_PLATFORM or TARGET_PLATFORM
 
+  From Mathew Robinson:
+    - Update cache debug output to include cache hit rate.
+    - No longer unintentionally hide exceptions in Action.py
+
+  From Lukas Schrangl:
+    - Enable LaTeX scanner to find more than one include per line (PR #3403)
```
bdbaddog: Generally we omit the PR #, but include issue #'s if there is one for the fix.

luhk:
I removed the PR number from CHANGES.txt. I don't think there has been an issue for this to reference.

bdbaddog:
I"m not very familiar with LaTEX syntax. but does the comment before the regex not still apply? 
```
        # We have to include \n with the % we exclude from the first part
        # part of the regex because the expression is compiled with re.M.
        # Without the \n,  the ^ could match the beginning of a *previous*
        # line followed by one or more newline characters (i.e. blank
        # lines), interfering with a match on the next line.
        # add option for whitespace before the '[options]' or the '{filename}'
```
It seems like if you change the regex, minimally you should update the comment before it referring to the bit you changed? Or am I misreading the comment/regex?

luhk:
Oops, I missed that. So it seems like checking for `\n` is necessary in combination with checking for `%`, otherwise a possible match may not be found because there was a comment in a previous line. But as I said, I don't think checking for comments is necessary since they are stripped in `canonical_text()` anyways. Therefore I removed the (Python source code) comment.

PR #3410, opened 2019-07-19
[WIP] Improve performance of Executor.get_lvals by removing delayed loading/caching

grossag:
This is another integration of an old VMware change. Original change description:

The Executor class has a get_lvars() method, which returns a dictionary
of helper TSList and TSObject instances, which call various methods on
Executor to get source and target lists. Comments mention that they
were added to delay creation of NodeList objects until needed, but the
effect is the opposite - the lists are re-created just to get a single
element or even attributes of the list objects. Another potential issue
is that the returned list can be mutated by adding sources/targets.

Profiling shows that this is a huge bottleneck for null or incremental
Fusion builds because of repeated get_subst_proxy() and list creation.
This change completely removes TSList and TSObject, and simplifies the
get_lvars() method to build the node lists from the current sources
and targets, just like other methods do.

With the change, a null build of Fusion's 'player' target is reduced
from 150 to 74 seconds, or 110 to 32 seconds for interactive builds.
The --profile pstats output shows substantially fewer calls:
   Function           Old # Calls    New # Calls
   get_lvars               10,250     10,250
   _get_targets            44,732          0
   get_subst_proxy    191,835,290    126,216

ESX builds (esx-all) don't show as much time improvement because there
is a lot of noise in other parts of the build. The calls have reduced:
   Function           Old # Calls    New # Calls
   get_lvars               61,959     61,959
   _get_targets           144,272          0
   get_subst_proxy     13,789,472    413,515

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/Executor.py
+++ b/src/engine/SCons/Executor.py
@@ -50,37 +50,6 @@ class Batch(object):
         self.targets = targets
         self.sources = sources
 
-
-
-class TSList(collections.UserList):
-    """A class that implements $TARGETS or $SOURCES expansions by wrapping
-    an executor Method.  This class is used in the Executor.lvars()
-    to delay creation of NodeList objects until they're needed.
-
-    Note that we subclass collections.UserList purely so that the
-    is_Sequence() function will identify an object of this class as
-    a list during variable expansion.  We're not really using any
-    collections.UserList methods in practice.
-    """
-    def __init__(self, func):
-        self.func = func
-    def __getattr__(self, attr):
-        nl = self.func()
-        return getattr(nl, attr)
-    def __getitem__(self, i):
-        nl = self.func()
-        return nl[i]
-    def __getslice__(self, i, j):
-        nl = self.func()
-        i, j = max(i, 0), max(j, 0)
-        return nl[i:j]
-    def __str__(self):
-        nl = self.func()
-        return str(nl)
-    def __repr__(self):
-        nl = self.func()
-        return repr(nl)
-
 class TSObject(object):
     """A class that implements $TARGET or $SOURCE expansions by wrapping
     an Executor method.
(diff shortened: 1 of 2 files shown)


acmorrow:
I was really excited to ty this out because I very much want faster null/no-op builds for MongoDB as well. But when I tried it (non-interactive), I found it was actually slower for me by about 10%:

master no-op rebuild:
Total build time: 98.272712 seconds
Total SConscript file execution time: 12.849968 seconds
Total SCons execution time: 85.422744 seconds
Total command execution time: 0.000000 seconds

real    1m40.153s
user    1m38.306s
sys     0m1.895s



patch no-op rebuild:
Total build time: 108.602506 seconds
Total SConscript file execution time: 12.687175 seconds
Total SCons execution time: 95.915330 seconds
Total command execution time: 0.000000 seconds

real    1m50.393s
user    1m48.421s
sys     0m2.044s

I wonder what is different about our builds.

bdbaddog:
> I was really excited to ty this out because I very much want faster null/no-op builds for MongoDB as well. But when I tried it (non-interactive), I found it was actually slower for me by about 10%:
> 
> master no-op rebuild:
> Total build time: 98.272712 seconds
> Total SConscript file execution time: 12.849968 seconds
> Total SCons execution time: 85.422744 seconds
> Total command execution time: 0.000000 seconds
> 
> real 1m40.153s
> user 1m38.306s
> sys 0m1.895s
> 
> patch no-op rebuild:
> Total build time: 108.602506 seconds
> Total SConscript file execution time: 12.687175 seconds
> Total SCons execution time: 95.915330 seconds
> Total command execution time: 0.000000 seconds
> 
> real 1m50.393s
> user 1m48.421s
> sys 0m2.044s
> 
> I wonder what is different about our builds.

They're internal SCons is based on 2.0.1 (or is it 2.1.0), that might explain such? (The patches don't help current, but helped old scons version?)

grossag:
The code in this class is very similar between 2.0.1 so it is possible that our builds are just a bit different in profile. I will retry the change internally on top of 3.0 when that is ready in a few months.

In the meantime we can close this and I will open a new pull request with just the test. The test of get_lvars would still be useful to have without the rest of this change,

PR #3420, opened 2019-08-04
[WIP] Add some debug logic to see what's going on with swig failures

bdbaddog:
Add some debug logic to see what's going on with swig failures


bdbaddog:
Not needed. Used Appveyor REST API to flush build cache.

PR #3421, opened 2019-08-04
Run latex after bibtex/biber only when necessary

luhk:
Although comments in src/engine/SCons/Tool/tex.py indicated that latex
should only be run after biber/bibtex if the .bbl file had changed, it
was always run.

## Contributor Checklist:

* [x] No documentation to update

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1066,6 +1066,9 @@ RELEASE 3.1.1 - Mon, 07 Aug 2019 20:09:12 -0500
         - JSON encoding errors for CacheDir config
         - JSON decoding errors for CacheDir config
 
+  From Lukas Schrangl:
+    - Run LaTeX after biber/bibtex only if necessary
+
 
 RELEASE 3.1.0 - Mon, 20 Jul 2019 16:59:23 -0700
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -63,6 +63,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
   From Ryan Saunders:
     - Fixed runtest.py failure on Windows caused by excessive escaping of the path to python.exe.
 
+  From Lukas Schrangl:
+    - Run LaTeX after biber/bibtex only if necessary
+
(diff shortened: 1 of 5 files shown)


mwichmann:
The AppVeyor fails are "known" and not caused by this PR: the swig-on-2.7 one is under investigation; the other is intermittent.

bdbaddog:
@managan  - Any chance you could take a look at this?

managan:
Not right now. Headed off for a week of vacation and a week long meeting in Italy.

I took a brief look. Right now we start by running latex and then decide whether to run again. We would have to store info about files that exist before the run so we could compare after to see if we need to run again.

Right now it just scans produced files to see if there are bibliography data present and then flags the files as read so it does not repeat that. So yes, if there is a bibliography we run latex at least twice.

I will try and see what I can do after I get back.

...

bdbaddog:
>PS. How do I search the issues list on github? Or is there a better place? 

github!

mwichmann:
https://help.github.com/en/articles/using-search-to-filter-issues-and-pull-requests

bdbaddog:
@managan  - Any chance you have a bit to review this PR now?

managan:
This looks good to me. 

I will note that if we want to support packages that create multiple bbl files, like the chapterbib or multibbl packages, then this will need a rewrite. The current logic assumes a single bbl file that is named from the primary text file name.

bdbaddog:
thanks @managan - So if we wanted to support packages that create multiple bbl files does the current logic (not inclusive of this patch) work?  in other words are we losing the capability by merging this?

managan:
OK,

It seems that the current system is conservative and works with packages like chapterbib. It does not clean all the files. But since it forces a run of bibtex on all the instances found in the aux file it catches changes.

So we do sort of lose capability.

Got to run so I can fill in details later.

bdbaddog:
@managan - Does it look like the code could be modified to avoid losing current functionality?

bdbaddog:
@managan - Ping? Any further thoughts?

PR #3427, opened 2019-08-19
Improve threading performance by sharing NodeInfo across threads

chasinglogic:
This is mildly work in progress but is based on results found [here](https://gist.github.com/chasinglogic/d43a1e12d43a7d8ca052fe2aa76f16a1)

The last section [Initializing NodeInfo at SCons.Node.FS.Base construction time](https://gist.github.com/chasinglogic/d43a1e12d43a7d8ca052fe2aa76f16a1) contains the relevant performance results.

The patch shown in that section however had some major flaws and this version maintains the same performance characteristics without side effects like loading SConsign before the SConstruct etc.

The observed behavior is that during a build all of the File nodes are passed out to threads, a single File node can be seen as a target or source multiple times. For instance one of our unit test binaries shows up 8 times in the DAG walk. The File node is shared across all threads and asked for it's `csig`. When calculating the `csig` the `NodeInfo` object is constructed because it does not exist. This object construction happens in each thread and so the `NodeInfo` is not actually shared across threads. I believe this is because of the use of `__slots__` since the `File` node does not see that it has a `self.ninfo` attribute but it's (multiple levels up) super class `Node` does.

Either way the content signature is calculated using a very expensive read and hash method then stored as the `csig` attribute on the `NodeInfo` object. This causes a long thread lock because the `File` node is the caller so is "doing the work". But since the `NodeInfo` object is not shared when the next thread sees this same `File` node it will cause the same long thread lock and do the same work again. This can repeat many many times.

You can see in the profile results in the above links that we go from spending `31.537` seconds locking threads and `22.198` on reading files to `9.494` seconds locking threads and `11.606` seconds reading files. Perhaps the biggest gains are on what we save with hashing. Time spent in `_hashlib.HASH` sees a reduction from `18.278` to less than 1 second:

```
    40289    0.689    0.000    0.689    0.000 {method 'update' of '_hashlib.HASH' objects}
```

I've seen thread locking times get as low as `4.738` seconds with some other SCons options like `--implicit-cache` and `--max-drift` settings set as according to the wiki.

When combined with the Subst rewrite in #3107 MongoDB builds could be sped up dramatically.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -6,9 +6,11 @@
 
 RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
-  From John Doe:
+  From Mathew Robinson:
 
-    - Whatever John Doe did.
+    - Improved threading performance by ensuring NodeInfo is shared
+      across threads. Results in ~13% improvement for parallel builds
+      (-j# > 1) with many shared nodes.
 
 
 RELEASE 3.1.1 - Mon, 07 Aug 2019 20:09:12 -0500
--- a/src/engine/SCons/Node/__init__.py
+++ b/src/engine/SCons/Node/__init__.py
@@ -614,6 +614,7 @@ class Node(object, with_metaclass(NoSlotsPyPy)):
         self._func_rexists = 1
         self._func_get_contents = 0
         self._func_target_from_source = 0
+        self.ninfo = None
 
         self.clear_memoized_values()
 
@@ -1131,11 +1132,10 @@ class Node(object, with_metaclass(NoSlotsPyPy)):
         return ninfo
 
     def get_ninfo(self):
-        try:
+        if self.ninfo is not None:
             return self.ninfo
-        except AttributeError:
-            self.ninfo = self.new_ninfo()
-            return self.ninfo
+        self.ninfo = self.new_ninfo()
+        return self.ninfo
 
     def new_binfo(self):
         binfo = self.BuildInfo()
--- a/src/engine/SCons/Node/FSTests.py
+++ b/src/engine/SCons/Node/FSTests.py
@@ -2545,11 +2545,10 @@ class FileTestCase(_tempdirTestCase):
 
             def get_ninfo(self):
                 """ mocked to ensure csig will equal the filename"""
-                try:
+                if self.ninfo is not None:
                     return self.ninfo
-                except AttributeError:
-                    self.ninfo = FakeNodeInfo(self.name, self.timestamp)
-                    return self.ninfo
+                self.ninfo = FakeNodeInfo(self.name, self.timestamp)
+                return self.ninfo
 
             def get_csig(self):
                 ninfo = self.get_ninfo()


src/engine/SCons/Node/__init__.py
```diff
@@ -1131,11 +1132,10 @@ def new_ninfo(self):
         return ninfo
 
     def get_ninfo(self):
-        try:
-            return self.ninfo
-        except AttributeError:
-            self.ninfo = self.new_ninfo()
+        if self.ninfo is not None:
```
bdbaddog, line 1135: pretty sure you can get away with "If self.ninfo" ? None equates to false and any false here would be an isssue?

chasinglogic: This is based on [PEP8's Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

I will change it if you want.

bdbaddog: Gotcha. I'll defer to PEP8 here.

chasinglogic:
@bdbaddog 

What's the best way to proceed? The appveyor failures look unrelated?

```
199/297 (67.00%) c:\\python27\\python.exe -tt test\Interactive\configure.py
match_re: mismatch at line 1:
  search re='^scons>>> .*foo\.cpp.*$'
  line='scons>>> scons: `foo.obj' is up to date.'
1,2c1
< scons>>> .*foo\.cpp.*
< scons>>> .*foo\.cpp.*
---
> scons>>> c:\python27\python.exe mycc.py foo.obj foo.cpp
3a3
> scons>>> c:\python27\python.exe mycc.py foo.obj foo.cpp
5c5
< scons>>>\s*
---
> scons>>> 
	at line 615 of C:\projects\scons\testing\framework\TestCommon.py (_complete)
	from line 675 of C:\projects\scons\testing\framework\TestCommon.py (finish)
	from line 117 of test\Interactive\configure.py
202/297 (68.01%) c:\\python27\\python.exe -tt test\Interactive\help.py
```


```
61/297 (20.54%) C:\\Python37\\python.exe test\Parallel\failed-build.py
Unexpected files exist: `C:\Users\appveyor\AppData\Local\Temp\1\testcmd.1232.fxqw2kok\f5'
	at line 555 of C:\projects\scons\testing\framework\TestCommon.py (must_not_exist)
	from line 122 of test\Parallel\failed-build.py
56/297 (18.86%) C:\\Python37\\python.exe test\NodeOps.py
```

I still need to update the src/CHANGES.txt and will do so. Not sure if this needs / how I would test this?

bdbaddog:
Yes Update CHANGES.txt with something like internal optimization should speed up checking if files have changed.  You can check the documentation and tests.  Current tests should cover functionality impacted by the change so you can check that one as well.

You can ignore those windows failures as they are intermittently failing.

chasinglogic:
Done. Thanks!

src/CHANGES.txt
```diff
@@ -6,9 +6,11 @@
 
 RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
-  From John Doe:
+  From Mathew Robinson:
 
-    - Whatever John Doe did.
+    - Improved threading performance by ensuring NodeInfo is shared
```
bdbaddog, line 11: Hmm. maybe add for Parallel builds (using -j # > 1) or something to indicate when this applies?

chasinglogic: Done.

PR #3458, opened 2019-10-02
Fix a syntax error in the documentation

theoratkin:
Quotation marks do not match, which is a syntax error.

## Contributor Checklist:


bdbaddog:
thanks!

PR #3460, opened 2019-10-04
fix CheckFunc detection code for MSVC

jw0k:
This patch fixes detection of some functions on Visual 2019 with optimizations (`/O2`) enabled.

For example: `memmove` is incorrectly detected as not available on Visual 2019 with `/O2` enabled because `/O2` enables `/Oi` which replaces `memmove` with a compiler intrinsic, which in turn causes the test code fail to compile. The `#pragma function` prevents replacing the tested function with a compiler intrinsic.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -18,6 +18,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Turn previously deprecated debug options into failures:
       --debug=tree, --debug=dtree, --debug=stree, --debug=nomemoizer.
 
+  From Jacek Kuczera:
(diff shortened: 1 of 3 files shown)


bdbaddog:
Can you add a CheckFunc('memmove') to test/Configure/config-h.py test, also add a blurb to src/CHANGES.txt to describe what you fixed.

jw0k:
@bdbaddog Done.

bdbaddog:
@jw0k - Pending completion of CI builds. Looks good to me. I'll merge once they're done, unless there's a surprise. Thanks for the good work!

bdbaddog:
Merging. No test failures caused by this change.

jw0k:
@bdbaddog Thank you.

Issue #3469, opened 2019-10-28
SConf subsystem's conftest file naming mechanism is non-deterministic


acmorrow:
**Describe the bug**
If a command line option can cause a different sequence of configure checks to run, or to run in a different order, SCons can become confused about which tests are cached, or about which actions are associated with a given numbered configure check. This can lead to either re-running configure checks that should have still been cached, only because the numbering changed, or, in worse cases, to a misconfigured build. See https://jira.mongodb.org/browse/SERVER-23208 for an example.

**Required information**
* Link to SCons Users thread discussing your issue.
N/A, discussed on slack.

* Version of SCons
3.1.1, but probably applies to older versions as well.

* Version of Python
3.7

* Which python distribution if applicable (python.org, cygwin, anaconda, macports, brew,etc)
N/A

* How you installed SCons
Vendored into MongoDB repo

* What Platform are you on? (Linux/Windows and which version)
Linux/MacOS/Windows, multiple versions

* How to reproduce your issue?  Please include a small self contained reproducer. Likely a SConstruct should do for most issues.
See repro instructions in linked MongoDB ticket.

* How you invoke scons (The command line you're using "scons --flags some_arguments")
See above.

acmorrow:
The fundamental issue here is the global incrementing counter. I have an in-flight solution that uses the hash of the test body to differentiate configure tests. It isn't perfect, but I'll put up my WIP code in a PR that references this issue.

bdbaddog:
This fixed by merged Pull Request #3543

PR #3470, opened 2019-10-28
[WIP] Make conftests generated and/or built files have unique names - Issue#3469

acmorrow:
Address issue #3469 

> If a command line option can cause a different sequence of configure checks to run, or to run in a different order, SCons can become confused about which tests are cached, or about which actions are associated with a given numbered configure check. This can lead to either re-running configure checks that should have still been cached, only because the numbering changed, or, in worse cases, to a misconfigured build. See https://jira.mongodb.org/browse/SERVER-23208 for an example.

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/SConf.py
+++ b/src/engine/SCons/SConf.py
@@ -56,6 +56,7 @@ import SCons.Warnings
 import SCons.Conftest
 
 from SCons.Debug import Trace
+from collections import defaultdict
 
 # Turn off the Conftest error logging
 SCons.Conftest.LogInputFiles = 0
@@ -98,7 +99,7 @@ def SetProgressDisplay(display):
 
 SConfFS = None
 
-_ac_build_counter = 0 # incremented, whenever TryBuild is called
+_ac_build_counter = defaultdict(int)
 _ac_config_logs = {}  # all config.log files created in this build
 _ac_config_hs   = {}  # all config.h files created in this build
 sconf_global = None   # current sconf object
@@ -592,8 +593,30 @@ class SConfBase(object):
             raise SCons.Errors.UserError('Missing SPAWN construction variable.')
 
         nodesToBeBuilt = []
+        sourcetext = self.env.Value(text)
+        f = "conftest"
 
-        f = "conftest_" + str(_ac_build_counter)
+        if text is not None:
+            textSig = SCons.Util.MD5signature(sourcetext)
+            textSigCounter = str(_ac_build_counter[textSig])
+            _ac_build_counter[textSig] += 1
+
+            f = "_".join([f, textSig, textSigCounter])
+            textFile = self.confdir.File(f + extension)
+            textFileNode = self.env.SConfSourceBuilder(target=textFile,
+                                                       source=sourcetext)
+            nodesToBeBuilt.extend(textFileNode)
+
+            source = textFile
+            target = textFile.File(f + "SConfActionsContentDummyTarget")
+        else:
+            source = None
+            target = None
+
+        action = builder.builder.action.get_contents(target=target, source=[source], env=self.env)
+        actionsig = str(SCons.Util.MD5signature(action))
+        f = "_".join([f, actionsig])
+
         pref = self.env.subst( builder.builder.prefix )
         suff = self.env.subst( builder.builder.suffix )
         target = self.confdir.File(pref + f + suff)
@@ -602,16 +625,6 @@ class SConfBase(object):
             # Slide our wrapper into the construction environment as
             # the SPAWN function.
             self.env['SPAWN'] = self.pspawn_wrapper
-            sourcetext = self.env.Value(text)
-
-            if text is not None:
-                textFile = self.confdir.File(f + extension)
-                textFileNode = self.env.SConfSourceBuilder(target=textFile,
-                                                           source=sourcetext)
-                nodesToBeBuilt.extend(textFileNode)
-                source = textFileNode
-            else:
-                source = None
 
             nodes = builder(target = target, source = source)
             if not SCons.Util.is_List(nodes):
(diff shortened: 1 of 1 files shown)


acmorrow:
This is some WIP work I had for #3469. I'll leave some inline comments explaining in more detail.

src/engine/SCons/SConf.py
```diff
@@ -592,8 +593,30 @@ def TryBuild(self, builder, text = None, extension = ""):
             raise SCons.Errors.UserError('Missing SPAWN construction variable.')
 
         nodesToBeBuilt = []
+        sourcetext = self.env.Value(text)
+        f = "conftest"
+
+        if text is not None:
+            textSig = SCons.Util.MD5signature(sourcetext)
+            textSigCounter = str(_ac_build_counter[textSig])
+            _ac_build_counter[textSig] += 1
```
acmorrow, line 602: I'd been hoping to entirely eliminate the counter. But without it, you end up with many configure tests which have the same text body (notably, the empty body). If those configure tests came from different environments, you get the dreaded "scons: warning: Two different environments were specified for target" issue. So, we re-purpose the counter to differentiate identical files. That is sort of unfortunate, but it still works, because if the numbering ever changes in a "bad" way, it shouldn't matter: the sources are always identical.

src/engine/SCons/SConf.py
```diff
@@ -592,8 +593,30 @@ def TryBuild(self, builder, text = None, extension = ""):
...
+
+            f = "_".join([f, textSig, textSigCounter])
+            textFile = self.confdir.File(f + extension)
+            textFileNode = self.env.SConfSourceBuilder(target=textFile,
+                                                       source=sourcetext)
+            nodesToBeBuilt.extend(textFileNode)
+
+            source = textFile
+            target = textFile.File(f + "SConfActionsContentDummyTarget")
```
acmorrow, line 611: This is unfortunate, but without it, we don't have a real `Node` to pass as the `target` argument for the `get_contents` call below, and if there are any emitters in play that want to actually examine the targets, they will fall over. I'm open to suggestions on the name or other approaches to ensure that there is a real `Node` available for use as the `target`.

src/engine/SCons/SConf.py
```diff
@@ -592,8 +593,30 @@ def TryBuild(self, builder, text = None, extension = ""):
...
+        else:
+            source = None
```
acmorrow, line 613: I haven't explored whether this branch ever happens, or how to disambiguate if it does.

src/engine/SCons/SConf.py
```diff
@@ -592,8 +593,30 @@ def TryBuild(self, builder, text = None, extension = ""):
...
+            target = None
+
+        action = builder.builder.action.get_contents(target=target, source=[source], env=self.env)
```
acmorrow, line 616: Originally, what I wanted here was to synthesize this action string *before* making the source file, and appending that to the source file name. Then everything would just flow nicely. But I can't do that, because I need the name of the source file to feed in.

acmorrow:
@bdbaddog - Any thoughts on this proposed change? It solves a real problem for my users. We have optional modules that can be enabled for the build, and they introduce new configure tests. When those modules are subsequently disabled or opted-out, the configure tests can get confused. This leads to people putting `--config=force` in their `SCONSFLAGS`, which leads to a poor user experience because the configure tests can never be cached.

acmorrow:
@bdbaddog - I've rebased this onto master and force updated the branch to get a new test run. I took a look at some of the old test failures, and it seems that most of them are caused by the fact that the name of configuration test files are no longer predictable. Many tests are looking for `conftest_0` and failing because that file now has a hash that can't be known ahead of time attached. Any suggestions on how to approach that?

bdbaddog:
@acmorrow - the file names should still be deterministic right? So I'd say figure out what the file name should be going forward and replace conftest_0 with that?

bdbaddog:
@acmorrow - hop on Discord if you want to discuss..

src/engine/SCons/SConf.py
bdbaddog: Perhaps the answer then not to have different source filenames, but different target file names. This would avoid `"scons: warning: Two different environments were specified for target"`

acmorrow: @bdbaddog - We collide on the `target` nodes of `SConfSourceBuilder`, so that is what this is actually doing: producing different target names.

bdbaddog:
Some notes on this PR:

1. The PR does improve the current state of configure logic when tests may be run or not run or run in different order on different runs. There were minimally cases where the results should have been cached from a previous run and were rerun. And maximally cases where different runs might reuse incorrect object files to link, yielding false negatives or positives.  In those cases a rerun with ```--config=force``` and/or cleaning the configure temp file directory would repair this issue. An example false failure can occur when there's a .o generated to check for compile which has no main and thus when link is attempted will fail and the check will (possibly) fail incorrectly
2. Even with the logic in this PR there are a couple corner cases where a false positive or negative can occur (though once again cleaning the configure output files/directory and/or ```--config=force``` would resolve them).  Generally these corner cases fall into multi step builders.  Such as Program which is called via TryLink (among other checkers).  Where the source file is named according to it's content hash and the link action hash.  But there's no uniqueness provided against the expanded CCCOM. So two calls with different environments where the CCCOM is different, but not the check being called or the LINKCOM could yield a file naming conflict and a false positive or negative on the configure check.
3. It may be possible to get rid of the file naming counter if the ```SConfSourceBuilder()``` can be called either from a single Environment() (not from the env passed to Configure) or from DefaultEnvironment().

That said the improvements in this PR are worthy of merging. 
Per discussion with @acmorrow I'll be pulling the logic from this PR and adding some tests and submitting a new PR.

Additionally I'll try to quantify with some tests item 2 above where a corner case might fail this logic and create an issue for that.

bdbaddog:
Work continued in PR #3543

PR #3473, opened 2019-11-06
Fix incorrect testing assumptions for Solaris

kulikjak:
As promised I tried to fix as many issues in scons on Solaris as I could. So far I managed to fix 26 errors with 41 still remaining (and I do plan to fix those as well).

The main issue was inconsistent use of file extensions in different parts of scons - I changed all i found to .pic.o (not sure why this one is used for Solaris, but it doesn't matter as long as it is used everywhere).

Most of those remaining issues are (most likely) due to scons using non-GNU, Solaris specific programs (like ar, tar, and possibly others), which can be different. This can often be solved by changing PATH to /usr/gnu/bin:$PATH to load GNU variants but that has no effect on scons as far as I can tell.

Is there a way to change the path globally for Solaris? I found that this can be done in SConstruct files, but that doesn't help us much here...

I still have to investigate relocation errors, not sure why that is happening.

I am also attaching log of still failing tests for reference.
[scons-failing.log](https://github.com/SCons/scons/files/3813562/scons-failing.log)


src/engine/SCons/Tool/MSCommon/vc.py
```diff
@@ -161,11 +161,18 @@ def get_host_target(env):
     if not host_platform:
         host_platform = platform.machine()
 
+    # Solaris returns i86pc for both 32 and 64 bit architectures
```
bdbaddog, line 164: Why does this matter? Does it break some tests?

datalogics-robb:
These changes do look reasonable to me, although I'm not familiar with the test suite.
The change to vc.py might be unnecessary, as we wouldn't have Visual Studio on Solaris, but I can verify that the architecture reported on Intel solaris differs from that on Intel Linux, and values such as 'sunos', 'sunos5', etc. are identifiers used under Solaris.

bdbaddog:
Please add a blurb to src/CHANGES.txt and then I'll merge.

src/engine/SCons/Tool/MSCommon/vc.py
kulikjak: All platforms are distinguishing between 32 and 64 bits, however, getting this info from `host_platform` and `_ARCH_TO_CANONICAL` is not possible on Solaris because `host_platform` itself is not enough to determine the architecture. That is why the canonical platform is being determined this way.

kulikjak:
Thanks for merge as this fixes many Solaris related problems, but there are still some test failing due to the usage of non-GNU Solaris specific programs (as mentioned in the description of this pull request).

Is there a way to change the default path globally for Solaris (to prioritize /usr/gnu/bin before /usr/bin)? I found that this can be done in SConstruct files, but that doesn't help much here as I would like to do so for anything being sconstructed on Solaris ...

bdbaddog:
Please bring the discussion to users mailing list or discord server or IRC channel?

kulikjak:
I see.

`/usr/gnu/bin` is not prioritized on default Solaris install.

I will investigate those tests more, see what more to do and let you know in the above-mentioned mailing list.

PR #3479, opened 2019-11-25
Add support for Value objects being implicit dependencies

grossag:
As part of consolidating outside dependencies, the code I work on takes on Value objects as implicit dependencies. That allows us to take on a dependency to an entire build (e.g. our compiler) rather than 500 files from it.

This has worked fine in practice for months now, but it turns out to break when using caching, because Node.get_contents() expects all dependencies to have the "name" attribute. This change adds that attribute to the Value class and a test to confirm that Node.get_contents() works now.

## Contributor Checklist:


bdbaddog:
So having a target with an implicit dependency on a Value node breaks being able to cache that target?

grossag:
Yeah that’s right @bdbaddog

bdbaddog:
o.k can you make a test which demonstrates that? (You've added a unit test, so this would be a system type test)

grossag:
> o.k can you make a test which demonstrates that? (You've added a unit test, so this would be a system type test)

Ok I'll give it a try. So far it's not obvious what this would look like and where it would go, so let me know if you have any guidance you want to provide. I found test\Value.py but this doesn't seem to fit there. I found test\CacheDir folder with lots of files in it; I may see if I can add something there.

grossag:
I'm putting this on hold for a little while so I can get a better and more representative test case. So far I haven't been able to find a good one. I'll either close or update this in the next few weeks.

abingham:
FWIW, I can confirm that this is an issue. I have some data structures I use as dependencies when building a directory, and scons fails when looking for `name` on these values. I worked around this by just adding a `name` attribute to my nodes, but this PR looks like a more proper solution.

> I'm putting this on hold for a little while so I can get a better and more representative test case.

I can't supply the code I'm seeing this in, but I'd hate to see this PR die for lack of a "representative" test case. What exactly are we looking for? It seems to me that if `Value` nodes are legitimate dependencies of `Dir` nodes, then a test case demonstrating that this relationship works doesn't require any more justification.

src/engine/SCons/Node/PythonTests.py
```diff
@@ -110,6 +110,17 @@ def test___init__(self):
...
+
+class ValueChildTestCase(unittest.TestCase):
+    def test___init__(self):
+        """Test support for a Value() being an implicit dependency of a Node"""
+        value = SCons.Node.Python.Value('v')
+        node = SCons.Node.Node()
+        node._func_get_contents = 2  # Pretend to be a Dir.
+        node.add_to_implicit([value])
+        contents = node.get_contents()
+        assert len(contents) > 0
```
bdbaddog: Can you add also `assert contents == '%s %s'%(value.get_csig(), value.name)`

grossag: Done, although the expected contents needs to have a \n at the end.

grossag:
I added a system test that also reproduced the issue without the fix. I also covered a few more cases where Value nodes are dependencies either directly or as grandchildren. That will ensure that we have good coverage of this in the future.

src/engine/SCons/Node/Python.py
```diff
@@ -96,6 +96,10 @@ def __init__(self, value, built_value=None):
         if built_value is not None:
             self.built_value = built_value
 
+        # Set a name so it can be a child of a node and not break
+        # its parent's implementation of Node.get_contents.
+        self.name = value
```
bdbaddog: which get_contents in Node/__init__.py references a python value nodes name?

grossag: @bdbaddog get_contents_dir uses it as a sorting key. See https://github.com/SCons/scons/blob/master/src/engine/SCons/Node/__init__.py#L216

src/engine/SCons/Node/Python.py
```diff
@@ -177,7 +184,7 @@ def get_csig(self, calc=None):
         return contents
 
 
-def ValueWithMemo(value, built_value=None):
+def ValueWithMemo(value, built_value=None, name=None):
     global _memo_lookup_map
```
bdbaddog, line 188: you need to pass name to the created Value().

grossag: Ah sorry, I missed that. I have added a test that was broken with the previous code and is fixed now.

grossag:
I fixed the issue you raised. I also updated the documentation; let me know if you think additional changes need to be made to the docs.

bdbaddog:
Merging. There's a few corners I'll address in follow up PR.
(caching value which has name != value where there are two Value nodes with different names)

PR #3481, opened 2019-11-26
Fix a problem under Solaris when using Python 3

datalogics-robb:
Fix a problem under Solaris when using Python 3, while maintaining support for Python 2.7.  subproces.Popen() produces bytes without the encoding argument in 3, which is not recognized by 2.7, and doesn't need to be decoded.

Also made a few changes to make "pycodestyle" happy for PEP8 compliance.

The issue is Solaris specific, and I don't believe it warrants a test without a Solaris test node.
Manually tested with Python 2.7 and 3.7 on Solaris/SPARC machines.

### DIFF
--- a/src/engine/SCons/Tool/suncxx.py
+++ b/src/engine/SCons/Tool/suncxx.py
@@ -73,7 +73,7 @@ def get_package_info(package_name, pkginfo, pkgchk):
             popen_args = {'stdout': subprocess.PIPE,
                           'stderr': DEVNULL}
             if PY3:
-                popen_args['encoding'] = 'utf-8'
+                popen_args['universal_newlines'] = True
             p = subprocess.Popen([pkginfo, '-l', package_name],
                                  **popen_args)
         except EnvironmentError:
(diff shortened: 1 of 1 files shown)


mwichmann:
Python 3.5 failed on Travis CI:
```
TypeError: __init__() got an unexpected keyword argument 'encoding':
...
 File "/home/travis/build/SCons/scons/src/engine/SCons/Tool/suncxx.py", line 78:
    **popen_args)
```

mwichmann:
Sigh...
> Changed in version 3.6: Added encoding and errors parameters

datalogics-robb:
@kulikjak - I realize now that I was getting this error in the scons-3.1.0 which does not have your recent change to add "decode()" - so in your opinion, is this change unnecessary?

bdbaddog:
Tests passing. Merging.

bdbaddog:
Oops forgot to ask for a blurb for CHANGES.txt
Please respond with one here and I'll add it and check it into master.

datalogics-robb:
Handle Unicode differences in python 2 & 3 when checking for Oracle Studio compiler.

bdbaddog:
How's this:

```
  From Rob Boehne
    - Fix suncxx tool (Oracle Studio compiler) when using Python 3.  Previously would throw an exception.
      Resolved by properly handling tool version string output as unicode.
```

datalogics-robb:
@bdbaddog looks good to me.

bdbaddog:
o.k. done. Thanks!

mwichmann:
@datalogics-robb  not to pile on or anything, but are you in any position to cross-check any of the work in #3473?  Since Solaris users are thin on the ground at the moment...

datalogics-robb:
What would you like me to do?

bdbaddog:
> What would you like me to do?

If you can take a look at the pull request #3473 and see if the changes look reasonable? Make any comments on the changes in that pull request. That would be very helpful.

PR #3483, opened 2019-12-02
Add support for scanning multiple entries in an action string

grossag:
This change adds support for scanning multiple entries in an action string in order to better support the following use cases:

1. A file is provided in an action string and should be taken as a dependency. For example, an action string "$PERL somefile.pl".
2. An action string actually has two actions separated by &&. For example, "cd <some_dir> && $ZIP <args>".

Adding support for #1 actually allows us to fix the test IMPLICIT_COMMAND_DEPENDENCIES.py on Windows, which was previously treating a Python file as executable even on Windows. This was causing tests to repeatedly open the default handler of Python files, which if set to Visual Studio causes DDE hangs. This test is fixed because now we can have the action string specify python as the first command and still take an implicit dependency on the script, which is now the second command.

## Contributor Checklist:


src/engine/SCons/Action.py
```diff
@@ -853,6 +853,16 @@ def __init__(self, cmd, **kw):
...
 
+        # Regex meant to parse a file path in the following cases:
+        #   /path/to/file
+        #   '/path/to/file'
+        #   "C:\path\to\file"
+        # Does not allow for stripping off an optional @ before the parameter.
+        # We don't need to support detecting temp files generated by
+        # TempFileMunge because in signature mode (which is what is used during
+        # action scanning) TempFileMunge doesn't modify the command string.
+        self.command_strip = re.compile('^[\'"]?([^\'"]*)[\'"]?$')
```
bdbaddog: Why not have this stack in the whole .py file..

src/engine/SCons/Action.py
```diff
@@ -998,6 +1008,44 @@ def get_implicit_deps(self, target, source, env, executor=None):
...
+                                        else ('&', '-')):
+                        m = strip_quotes.match(d)
+                        if m:
+                            d = m.group(1)
+
+                        # For now, only match files, not directories.
+                        p = os.path.abspath(d) if os.path.isfile(d) else \
+                            env.WhereIs(d)
+                        if p:
+                            res.append(env.fs.File(p))
```
bdbaddog: Why not just File()?

src/engine/SCons/Action.py
```diff
@@ -998,6 +1008,44 @@ def get_implicit_deps(self, target, source, env, executor=None):
...
+                        else:
+                            # Try to find the dependency in any source
```
bdbaddog: I think In general including path into repo is deprecated.  Should just look for non-repo file. There's a way to query if File is known without creating the node.

grossag: So are you saying the “else” case should be removed because repo support is not needed?

src/engine/SCons/Action.py
grossag: I don’t know. What would File() replace?

test/Batch/CHANGED_SOURCES.py
```diff
@@ -48,8 +48,11 @@
 sys.exit(0)
 """)
 
+# Disable IMPLICIT_COMMAND_DEPENDENCIES because otherwise it renders them less
```
bdbaddog: Can you explain? Not sure I understand.

src/CHANGES.txt
```diff
@@ -16,6 +16,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       is a reasonable default and also aligns with changes in Appveyor's VS2019 image.
     - Drop support for Python 2.7. SCons will be Python 3.5+ going forward.
 
+  From Adam Gross:
+    - Added support for scanning multiple entries in an action string if
+      implicit command dependency scanning is enabled.
```
bdbaddog: add how this is done and what the values mean

bdbaddog:
Good work.
Not sure why you're needing to turn off IMPLICIT_COMMAND_DEPENDENCIES for some tests.
Does this imply you've changed the default behavior?

Also right now looks like you have three options: no Implicit, only first argument, or all arguments.

I think it could be useful to be able to specify to process the first N arguments  and/or a list of which arguments.
Also needs a manpage entry update for new IMPLICIT_COMMAND_DEPENDENCIES values.

test/Batch/CHANGED_SOURCES.py
grossag: Sorry this is a remnant of my change previously switching the default behavior. This can be reverted now that the default behavior is unchanged. I'll make a pass through the diff reverting unnecessary changes.

grossag:
> Good work.
> Not sure why you're needing to turn off IMPLICIT_COMMAND_DEPENDENCIES for some tests.
> Does this imply you've changed the default behavior?
> 
> Also right now looks like you have three options: no Implicit, only first argument, or all arguments.
> 
> I think it could be useful to be able to specify to process the first N arguments and/or a list of which arguments.
> Also needs a manpage entry update for new IMPLICIT_COMMAND_DEPENDENCIES values.

Thanks for the feedback. Turning off IMPLICIT_COMMAND_DEPENDENCIES was a mistake, a remnant of my older change modifying the default behavior. I have reverted that change. I will also look at updating the documentation and CHANGES.txt.

I would prefer to avoid customizing the behavior any further. I understand that we don't want to change the current behavior of IMPLICIT_COMMAND_DEPENDENCIES and the default behavior. Three main reasons:

1. The testability overhead.
2. A lack of an intended consumer and clear use cases.
3. The variance of action strings making it difficult to find meaning in a specific value of N. For example, N=2 may make sense for a Perl script (e.g. N=1 taking a dependency on Perl binary and N=2 taking a dependency on the script path) but not for a linker action.

bdbaddog:
Since you can set.
`env.Command(....., IMPLICIT_COMMAND_DEPENDENCIES=something which only make sense for this command)`
Then it can make sense to have it more customizable..

grossag:
In my opinion, if someone is going to go to the trouble of overriding IMPLICIT_COMMAND_DEPENDENCIES for a specific action, they should just use env.Depends instead. The benefit of implicit command dependency scanning is having a safety net for people not specifying their dependencies correctly. Specifying dependencies manually is always preferred, but this page is useful for large codebases where people aren’t always doing it correctly.

That said, if you won’t accept the PR without this change then I will do it. So I can better understand what you are asking for, here is what I understand you to be requesting:

IMPLICIT_COMMAND_DEPENDENCIES value possibilities:

0/False: no implicit dependencies (as before)
1/True: only 1 implicit dependency even if && is in the command string (as before)
Another integer value N: N implicit dependencies before a && (but what about after? N deps after each && too)
“all”: all implicit dependencies before and after &&

bdbaddog:
Since values in Command can vary by environment and/or overrides and those values may need to be added to implicit dependencies, allowing more control over the implicit scanning of command line can provide value.

I was thinking:
IMPLICIT_COMMAND_DEPENDENCIES value possibilities:

```
0/False: no implicit dependencies (as before)
1/True: only 1 implicit dependency even if && is in the command string (as before)
Another integer value N: N implicit dependencies before a && (but what about after? N deps after each && too)
“all”: all implicit dependencies before and after &&
A list with position values: implicit dependencies only for those positions listed.
```

Seems like you're already 95% of the way to supporting a list.
But if that's not interesting to you, I'm not going to hold up the PR for u.

IMHO.. if you need to use env.Depends() you've done something wrong and/or SCons needs to do more. (for 90% of the cases)

grossag:
I have updated the PR with support for `IMPLICIT_COMMAND_DEPENDENCIES` being an integer.

PR #3486, opened 2019-12-06
Fix multithreaded Windows builds when a thread has a file open for write

grossag:
Fix multithreaded Windows builds when a thread has a file open for write

Python 2 enables handle inheritance by default for child processes. It wasn't until Python 3.4 that it was disabled. This causes problems because if a Python action is writing to a file and a child process is spawned at that exact moment, builds fail because of sharing issues.

## Contributor Checklist:

* [x] I have created a new test or updated the unit tests to cover the new/changed functionality. (Waived by @bdbaddog  )
* [x] I have updated the appropriate documentation (None - @bdbaddog )


grossag:
I see I broke src\engine\SCons\SConfTests.py. I'll look into that Monday.

bdbaddog:
Interesting so while I don't see the "N" modifier in python 2.7's open docs. If I dig in the source code python C source calls windows _wfopen() function which does document the "N" arg here.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019

CPython sanitizes the mode strings but explicitly ignores "N"  here:
https://github.com/python/cpython/blob/2.7/Objects/fileobject.c#L212

bdbaddog:
So likely the failure is due to this code in SConfTests.py
https://github.com/SCons/scons/blob/master/src/engine/SCons/SConfTests.py#L58

```
        for n in list(sys.modules.keys()):
            if n.split('.')[0] == 'SCons' and n[:12] != 'SCons.compat':
                m = sys.modules[n]
                if isinstance(m, ModuleType):
                    # if this is really a scons module, clear its namespace
                    del sys.modules[n]
                    sys.stderr.write("clearing :%s\n"%n)
                    m.__dict__.clear()

```

Where this test (and likely only this test) trys to reset all the SCons modules by deleting them from sys.modules (except for scons.compat)

bdbaddog:
Here's the stacktrace causing the platform overwrite of open for win32:

```
  File "src\engine\SCons\SConfTests.py", line 264, in test_TryRun
  File "src\engine\SCons\SConfTests.py", line 77, in _resetSConfState
  File "C:\Users\Sylvia\devel\grossagg\src\engine\SCons\Environment.py", line 947, in __init__
    platform = SCons.Platform.Platform()
  File "C:\Users\Sylvia\devel\grossagg\src\engine\SCons\Platform\__init__.py", line 288, in Platform
    module = platform_module(name)
  File "C:\Users\Sylvia\devel\grossagg\src\engine\SCons\Platform\__init__.py", line 105, in platform_module
    mod = importlib.import_module("." + name, __name__)
  File "C:\Python27\lib\importlib\__init__.py", line 37, in import_module
    __import__(name)
  File "C:\Users\Sylvia\devel\grossagg\src\engine\SCons\Platform\win32.py", line 99, in <module>
    __builtin__.file = _scons_file
  File "C:\Users\Sylvia\devel\grossagg\src\engine\SCons\Platform\win32.py", line 99, in <module>
    __builtin__.file = _scons_file
```

grossag:
Sorry I assumed it was documented and hadn’t looked far into it. Here is the summary from the initial commit a while back:

“Seeing that the Python built-ins just wrap the CRT's `fopen` function,
I checked the latter's docs and learned of the `N` mode option,
which tells Windows that the requested handle is not to be inherited
by child processes. The `N` option has been around since
at least MSVC 9.0, which our Python distribution is built against,
and so this change rewrites SCons's monkeypatched functions to forcibly
insert `N` into the `mode` arg passed to the Python built-ins.

This patch isn't a perfect fix, but it's ridiculously easy and gets
the job done.  For example, it's still possible to inherit handles from
directories or through any other constructor that
uses neither `file` nor `open`, but I still believe this change to be
a strong improvement.”

bdbaddog:
I played around with your fix and was able to get the tests to pass if I changed to this

```
        _builtin_open = open
        _builtin_file = file

        def _scons_fixup_mode(mode):
            """Adjust 'mode' to mark handle as non-inheritable.

            SCons is multithreaded, so allowing handles to be inherited by
            children opens us up to races, where (e.g.) processes spawned by
            the Taskmaster may inherit and retain references to files opened
            by other threads. This may lead to sharing violations and,
            ultimately, build failures.

            By including 'N' as part of fopen's 'mode' parameter, all file
            handles returned from these functions are atomically marked as
            non-inheritable.
            """
            if not mode:
                # Python's default is 'r'.
                # https://docs.python.org/2/library/functions.html#open
                mode = 'rN'
            elif 'N' not in mode:
                mode += 'N'
            return mode

        def _scons_file(name, mode=None, *args, **kwargs):
            global _builtin_file
            return _builtin_file(name, _scons_fixup_mode(mode), *args, **kwargs)

        def _scons_open(name, mode=None, *args, **kwargs):
            global _builtin_open
            return _builtin_open(name, _scons_fixup_mode(mode), *args, **kwargs)

        # import pdb; pdb.set_trace()
        file = _scons_file
        open = _scons_open
```

The main differences is not referencing/importing __builtin__, is that important to get the patch to work?

grossag:
@bdbaddog I confirmed that removing `__builtin__` causes the issue to come back. AFAICT just assigning to file causes the override to not be effective.

grossag:
Looks like I broke file-is-type-not-func.py because it specifically exists to deny the design I chose (replacing file with a function). I'll see whether I can replace file with a class instead of a function.

bdbaddog:
Merging

PR #3491, opened 2019-12-12
Optimize for most common case in Entry.disambiguate()

chasinglogic:
This removes many unnecessary os.stat and related FS IO calls. Saves about 2 seconds on MongoDB's no-op rebuild time.

## Contributor Checklist:


bdbaddog:
Did you see any perf improvement with this?

chasinglogic:
>  Saves about 2 seconds on MongoDB's no-op rebuild time.

Per original comment. Not much but it does help

PR #3504, opened 2019-12-18
Improve performance of Subst by preventing unnecessary frame allocations

chasinglogic:
This removes 5 - 6 seconds off Subst in the MongoDB build: ~13 seconds down from ~18 seconds.

These classes were never capturing any values from the functions but the Python interpreter doesn't know that so it won't cache anything about these classes. In this new code the class objects and related dictionaries are only ever allocated once.

## Contributor Checklist:


bdbaddog:
Merging.

PR #3518, opened 2020-01-06
Prevent unnecessary recursion when value is already expanded

chasinglogic:
I had a similar change for StringSubber but it provided little to no perf gain so I have neglected to add it. I can make that change again however if it's decided these should work as similarly as possible.

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/Subst.py
+++ b/src/engine/SCons/Subst.py
@@ -499,6 +499,9 @@ class ListSubber(collections.UserList):
         self.in_strip = None
         self.next_line()
 
+    def expanded(self, s):
+        return is_String(s) and _separate_args.findall(s) is None
+
     def expand(self, s, lvars, within_list):
         """Expand a single "token" as necessary, appending the
         expansion to the current result.
@@ -553,6 +556,10 @@ class ListSubber(collections.UserList):
                      raise_exception(NameError(), lvars['TARGETS'], old_s)
                 elif s is None:
                      return
+
+                if self.expanded(s):
+                    self.append(s)
+                    return
 
                 # Before re-expanding the result, handle
                 # recursive expansion by copying the local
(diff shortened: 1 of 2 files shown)


src/engine/SCons/Subst.py
```diff
@@ -493,6 +493,9 @@ def __init__(self, env, mode, conv, gvars):
         self.in_strip = None
         self.next_line()
 
+    def expanded(self, s):
+        return is_String(s) and _separate_args.findall(s) is None
```
bdbaddog: would "$" in s be sufficent?

chasinglogic: Not quite this can get strings that have spaces that wouldn't just be self.appended but would be split up for instance:

'foo something $bar'

It's also faster to just re-use the regex here since Python does a pretty good job of regex caching.

bdbaddog: Gotcha

src/engine/SCons/Subst.py
```diff
@@ -542,6 +549,10 @@ def expand(self, s, lvars, within_list):
                     else:
                         return
 
+                if self.expanded(s):
```
bdbaddog, line 574: Stick a comment here.

src/engine/SCons/Subst.py
```diff
@@ -493,6 +493,13 @@ def __init__(self, env, mode, conv, gvars):
         self.in_strip = None
         self.next_line()
 
+    def expanded(self, s):
+        if not is_String(s) or isinstance(s, CmdStringHolder):
```
bdbaddog, line 511: docstring please.

src/engine/SCons/Subst.py
chasinglogic: Done.

src/engine/SCons/Subst.py
chasinglogic: Done.

PR #3534, opened 2020-01-24
[WIP] Clear dangling links before building targets

### LINKED ISSUE #3516: SCons does not remove dangling symlink target before build
Hi

According to https://scons.org/doc/3.1.2/HTML/scons-user/ch13.html, _when SCons determines that an target file needs to be rebuilt and removes the existing version of the target before executing_.

Unfortunately the above is not true, when the target file for some reason is a symbolic link to non-existing file. In such a case scons leaves the broken symlink as is, and then runs target build action, which then might try to create target file in the destination of the symlink (for example g++ works like this).

Is there any possibility to convince scons to remove the target's dangling symlink before starting  build?

chasinglogic:
Closes #3516

Noticed this issue when working on my install-as-command branch.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -35,6 +35,7 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       unnecessary calls to eval when a token is surrounded in braces
       but is not a function call.
     - Improve performance of subst by removing unnecessary recursion.
+    - Cleanup dangling symlinks before running builders
 
   From Mats Wichmann:
     - Remove deprecated SourceCode
--- a/src/engine/SCons/Node/__init__.py
+++ b/src/engine/SCons/Node/__init__.py
@@ -137,6 +137,8 @@ def exists_always(node):
     return 1
 
 def exists_base(node):
+    if node.islink():
+        return node.lstat() is not None
     return node.stat() is not None
 
 def exists_entry(node):
--- a/src/engine/SCons/Node/__init__.py
+++ b/src/engine/SCons/Node/__init__.py
@@ -138,7 +138,7 @@ def exists_always(node):
 
 def exists_base(node):
     if node.islink():
-        return node.lstat() is not None
+        return node.lstat() is not None and node.stat() is not None
     return node.stat() is not None
 
 def exists_entry(node):
--- a/src/engine/SCons/Node/FS.py
+++ b/src/engine/SCons/Node/FS.py
@@ -3138,7 +3138,10 @@ class File(Base):
         SCons.Node.Node.prepare(self)
 
         if self.get_state() != SCons.Node.up_to_date:
-            if self.exists():
+            # Exists will report False for dangling symlinks so if it
+            # exists or is a link (which would mean it's a dangling
+            # link) then we should remove it as appropriate.
+            if self.exists() or self.islink():
                 if self.is_derived() and not self.precious:
                     self._rmv_existing()
             else:
--- a/src/engine/SCons/Node/FS.py
+++ b/src/engine/SCons/Node/FS.py
@@ -711,6 +711,20 @@ class Base(SCons.Node.Node):
         self._memo['stat'] = result
         return result
 
+    @SCons.Memoize.CountMethodCall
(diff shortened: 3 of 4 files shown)


chasinglogic:
@bdbaddog this has tests now

chasinglogic:
@bdbaddog The codecov failures I can't see? They just say "Unable to compare commits because the base of the compare did not upload a coverage report."

bdbaddog:
> @bdbaddog The codecov failures I can't see? They just say "Unable to compare commits because the base of the compare did not upload a coverage report."

I wouldn't worry about that. It doesn't gate the PR.
It's likely due to the issue with the .travis.yml

src/CHANGES.txt
```diff
@@ -35,6 +35,7 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       unnecessary calls to eval when a token is surrounded in braces
       but is not a function call.
     - Improve performance of subst by removing unnecessary recursion.
+    - Cleanup dangling symlinks before running builders
```
bdbaddog, line 38: Please add a note which Issue this resolves.

src/engine/SCons/Node/__init__.py
```diff
@@ -137,6 +137,8 @@ def exists_always(node):
     return 1
 
 def exists_base(node):
+    if node.islink():
```
bdbaddog, line 140: can we move this logic into FS.Base.stat()?

chasinglogic: You could but I would think that it would be weird that the `stat` method no longer corresponds to the stat syscall. It would then be doing something more akin to `exists` (which is eventually what calls this function anyway currently).

I'm happy to make that change if you think that's right

bdbaddog: Could this be simplifed ot skip the islink() like this:

```
def exists_base(node):
    return node.lstat() is not None and node.stat() is not None 
```

It looks like lstat() on a non symlink returns same info as stat() so worst case you effectively ask for stat twice?

From python 3.5 docs

> On platforms that do not support symbolic links, this is an alias for stat().
> 
> As of Python 3.3, this is equivalent to os.stat(path, dir_fd=dir_fd, follow_symlinks=False).

chasinglogic:
Closing this because it is unkown if/when I will get to addressing the test failures and code review comments.

bdbaddog:
I'm going to reopen and mark WIP. I"ll see if I can bring it across the finish line.

bdbaddog:
Please don't delete your branch.

acmorrow:
@bdbaddog - This seems like a worthwhile thing to fix. What needs to happen to get this issue moving again from where Mat left off?

acmorrow:
@bdbaddog - Should this be included for 4.0?

src/engine/SCons/Node/__init__.py
dmoody256: I don't think exists_base needs to be changed at all here based off a small table I made, you'll notice the results are the same for `(stat is not None)` and `(stat is not None and lstat is not None)`:

  | symlink exists, file exists | symlink exists, file doesnt exists | symlink doesn't exist, file exists | symlink doesn't exist, file doesnt exist
-- | -- | -- | -- | --
stat is not None | TRUE | FALSE | FALSE | FALSE
lstat is not None | TRUE | TRUE | FALSE | FALSE
stat is not None and lstat is not None | TRUE | FALSE | FALSE | FALSE

bdbaddog: @dmoody256  did you try linux, windows, macos?

dmoody256: I tested manually on linux, and for windows, lstat would become the same as stat according to the docs you quoted. In my follow on [PR](https://github.com/SCons/scons/pull/3638), the tests introduced in this PR ran and passed on windows and linux with no change to the exists_base function.

So really macos is the only one untested, and I am assuming its the same as other *nix platforms for symlinking. Someone correct me if I'm wrong.

Issue #1323, opened 2018-01-02
makefile generator builder


bdbaddog:
This issue was originally created at: 2006-01-26 14:36:46.
This issue was reported by: `veplaini`.
veplaini said at 2006-01-26 14:36:46
>I would like to suggest a Makefile builder. This would
>be useful for generating makefiles for systems without
>scons/python.
> 
>In a similar vein to the MSVC project builder

issues@scons said at 2006-01-26 14:36:46
>Converted from SourceForge tracker item 1415784

gregnoel said at 2008-08-12 15:31:26
>Bug party triage.  Contact Axel W. for his Tigris ID and assign it to him.

gregnoel said at 2008-08-17 16:21:31
>Reassign to Axel Wachtler, who is making some good progress on doing this.

gregnoel said at 2008-08-17 16:31:50
>Added myself as CC so I can see how this is going.

michaelis_l said at 2009-05-08 06:23:04
>Added myself as CC so I can see how this is going, too.

PR #3536, opened 2020-01-24
[WIP] write a tool to generate build.ninja files from SCons

chasinglogic:
**NOTE THIS PR IS CLOSED
The work has been moved to PR #3642 
I'm locking this PR to further comments. Please comment on #3642** 

This tool can generate ninja files from SCons, it is generic but does fall back to slow cases of executing scons from ninja when necessary (there are too many possible things SCons can do that don't translate into "ninja world")

We've tried to keep this file heavily commented since it reaches into SCons internals and has to handle edge cases both within SCons itself and in Ninja.

Recommended reading this Ninja build manual: https://ninja-build.org/manual.html

This tool requires the use of [ninja_syntax.py](https://github.com/ninja-build/ninja/blob/master/misc/ninja_syntax.py) and attempts to load it from an environment variable since the file is intended to be vendored. I wasn't sure if it should be vendored in SCons with the tool or left as an exercise for the user.

You can see it in use at MongoDB here in our SConstruct: https://github.com/mongodb/mongo/blob/master/SConstruct#L3756

Some of our tools reference the Ninja generator since they disable behavior when it's on so if you would like to poke around our site_tools directory you can see what that looks like. While it's not strictly necessary to do so it has allowed us to bring down generation times.

For reference this can generate a ninja file for MongoDB in 17 - 30 seconds.

## Contributor Checklist:

### DIFF
--- a/src/engine/SCons/Tool/ninja.py
+++ b/src/engine/SCons/Tool/ninja.py
@@ -1,0 +1,1396 @@
+# Copyright 2019 MongoDB Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Generate build.ninja files from SCons aliases."""
+
+import sys
+import os
+import importlib
+import io
+import shutil
+
+from threading import Lock
+from os.path import join as joinpath
+from os.path import splitext
+
+import SCons
+from SCons.Action import _string_from_cmd_list, get_default_ENV
+from SCons.Util import is_String, is_List
+from SCons.Script import COMMAND_LINE_TARGETS, LOADED_SCONSCRIPTS
+
+NINJA_SYNTAX = "NINJA_SYNTAX"
+NINJA_RULES = "__NINJA_CUSTOM_RULES"
+NINJA_POOLS = "__NINJA_CUSTOM_POOLS"
+NINJA_CUSTOM_HANDLERS = "__NINJA_CUSTOM_HANDLERS"
+NINJA_BUILD = "NINJA_BUILD"
+NINJA_WHEREIS_MEMO = {}
+NINJA_STAT_MEMO = {}
+MEMO_LOCK = Lock()
+
+__NINJA_RULE_MAPPING = {}
+
+# These are the types that get_command can do something with
+COMMAND_TYPES = (
+    SCons.Action.CommandAction,
+    SCons.Action.CommandGeneratorAction,
+)
+
+
+def _install_action_function(_env, node):
+    """Install files using the install or copy commands"""
+    return {
+        "outputs": get_outputs(node),
+        "rule": "INSTALL",
+        "pool": "install_pool",
+        "inputs": [get_path(src_file(s)) for s in node.sources],
+        "implicit": get_dependencies(node),
+    }
+
+
+def _lib_symlink_action_function(_env, node):
+    """Create shared object symlinks if any need to be created"""
+    symlinks = getattr(getattr(node, "attributes", None), "shliblinks", None)
+
+    if not symlinks or symlinks is None:
+        return None
+
+    outputs = [link.get_dir().rel_path(linktgt) for link, linktgt in symlinks]
+    inputs = [link.get_path() for link, _ in symlinks]
+
+    return {
+        "outputs": outputs,
+        "inputs": inputs,
+        "rule": "SYMLINK",
+        "implicit": get_dependencies(node),
+    }
+
+
+def is_valid_dependent_node(node):
+    """
+    Return True if node is not an alias or is an alias that has children
+
+    This prevents us from making phony targets that depend on other
+    phony targets that will never have an associated ninja build
+    target.
+
+    We also have to specify that it's an alias when doing the builder
+    check because some nodes (like src files) won't have builders but
+    are valid implicit dependencies.
+    """
+    return not isinstance(node, SCons.Node.Alias.Alias) or node.children()
+
+
+def alias_to_ninja_build(node):
+    """Convert an Alias node into a Ninja phony target"""
+    return {
+        "outputs": get_outputs(node),
+        "rule": "phony",
+        "implicit": [
+            get_path(n) for n in node.children() if is_valid_dependent_node(n)
+        ],
+    }
+
+
+def get_dependencies(node):
+    """Return a list of dependencies for node."""
+    return [get_path(src_file(child)) for child in node.children()]
+
+
+def get_inputs(node):
+    """Collect the Ninja inputs for node."""
+    executor = node.get_executor()
+    if executor is not None:
+        inputs = executor.get_all_sources()
+    else:
+        inputs = node.sources
+
+    inputs = [get_path(src_file(o)) for o in inputs]
+    return inputs
+
+
+def get_outputs(node):
+    """Collect the Ninja outputs for node."""
+    executor = node.get_executor()
+    if executor is not None:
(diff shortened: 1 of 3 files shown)


cirosantilli:
This is to me the most important feature scons could possibly implement, and the only reason people use CMake. Building the dependency tree of a complex project takes too long (3s?) and Make/Ninja are basically a cache for that graph.

It would also be interesting to see how https://el-tramo.be/scons2ninja/ compares.

Related for Makefile: https://github.com/SCons/scons/issues/1323

cirosantilli:
Can you also provide the exact commands to run this so I can try it out on [gem5](http://www.gem5.org/)?

chasinglogic:
@cirosantilli Due to lack of interest from maintainers and community members this PR is woefully out of date.

I would recommend vendoring the latest version directly from MongoDB: https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja_next.py

To enable that vendor it like any SCons tool then in your SConstruct (towards the end ideally) do `env.Tool('ninja_next')`. You can then build a ninja file with `scons $WHATEVER_FLAGS_AND_OPTIONS build.ninja`.

Note that once this tool is loaded SCons will not function as a build system (i.e. it can't build targets itself) which is why MongoDB flag guards it in our SConstruct. You can see how we use it [here](https://github.com/mongodb/mongo/blob/master/SConstruct#L3792).

Our latest version is much more robust and faster than the version from this PR.

Happy to answer any questions you have or look at any bugs. If you're using `FunctionAction`s extensively you'll need to register handlers like we do in the SConstruct link above.

chasinglogic:
As for how this compares to https://el-tramo.be/scons2ninja/

- We are fully generic, we don't care about what tools are enabled whereas the above will fail if it sees a tool which it doesn't recognize.
- We handle environment variables, if SCons sees something in `env['ENV']` that's different from the `os.environ` we will encode that in the ninja file.
- We fully support dynamic content and generated files. We don't require you to manually delete anything to rebuild it.
- We batch generated files which must be built with SCons into a single SCons call. This makes the ninja file much faster.

There is another SCons to ninja translator that web page links to for NaCl that has the following caveats:

* Some constructs don't work, so not all of small_tests passes under Ninja at the moment.  For example, tests that use temp files via MakeEmptyFile() don't work yet.

 * Header file dependencies don't work very well.  This still uses Scons's header file scanner (rather than GCC's -M options), so it doesn't detect changes to C/C++ files' #include dependencies on header files unless you re-run Scons to regenerate the Ninja file.

 * Ideally we'd record which *.scons files got read, so that we could automatically re-run Scons to regenerate the Ninja file if the *.scons files changed.

Our ninja generator solves all of these problems, our ninja file knows how to regenerate itself and when to. We track header file dependencies correctly. Finally, we support all constructs in so far as it's possible. Again SCons lets you do things with builders that simply don't translate to Ninja at all. For these cases we allow you to register custom handlers to perform the translation, or in the case that you simply want to ignore it you can use an override variable for a builder call `NINJA_SKIP=True` and we will just ignore it.

src/engine/SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1396 @@
+# Copyright 2019 MongoDB Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
bdbaddog, line 3: Can we get this assigned to SCons and using MIT license?

bdbaddog:
@chasinglogic - how is the PR out of date? Did we change some SCons logic that the PR is depending on?

src/engine/SCons/Tool/ninja.py
acmorrow: I can definitely reach out to our legal about getting this (and other MongoDB SCons tools) over to MIT.

acmorrow:
> @chasinglogic - how is the PR out of date? Did we change some SCons logic that the PR is depending on?

Well, we have made subsequent changes to our copy of the tool, so this PR doesn't really reflect the state of the art anymore. The most recent version is here: https://github.com/mongodb/mongo/blob/60aed37d906100af704bcb5c4cb8f6080e1540fe/site_scons/site_tools/ninja_next.py

Note that in our tree it is called `ninja_next.py` so that we can make changes there without affecting users who are currently using `ninja.py`. We plan to periodically roll `ninja_next.py` over to `ninja.py` but haven't done so yet.

chasinglogic:
Closing this because it is unkown if/when I will get to addressing the test failures and code review comments.

acmorrow:
@chasinglogic - I'll try to make an updated version of this sometime in April. Thanks for all your work on making SCons able to generate Ninja. I agree with @cirosantilli above that this feature is perhaps one of the more important things that the SCons project could invest in integrating. @bdbaddog, can we discuss how to approach rebooting this work once I'm available again next month?

bdbaddog:
Reopening. I don't want to loose this PR until we open another one. Please don't delete the source branch.

bdbaddog:
@acmorrow - yes the intent is to get this into 4.0.0.  One not this and the compilation database would be much easier to merge in if they could be changed to MIT license from Apache. I believe @chasinglogic had pushed that up to your legal?

acmorrow:
I'm not sure if that outreach to legal happened or not, but I've made a note to myself to pursue it when I'm back full time in April.

dmoody256:
+1 for how important this feature is. The company I currently work for has been converting existing SCons builds to CMake purely to gain speed of iterative ninja builds.

src/engine/SCons/Tool/ninja.py
acmorrow: FYI this relicense was completed: https://github.com/mongodb/mongo/commit/77162e6d10c23eb05e8805290ac8dc0b579f04a7

Also please note that the newest version of the MongoDB ninja tool is here: https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja_next.py

There are still a few small mongodb-isms in there, but we will be working to scrub them out. Some of the recent issues that I've filed are related to doing so.

acmorrow:
> Can you also provide the exact commands to run this so I can try it out on [gem5](http://www.gem5.org/)?

@cirosantilli - want to reach out on the SCons discord and I can discuss? The answer is too long and interactive to reasonably cover here.

dmoody256:
I pulled in ninja-next.py into my own branch of this and i'm writing some tests for it. As-is I think its mostly done and good enough, I'm still trying to come up with some edge cases to test though. 

As for the ninja-syntax.py requirement, somehow keep track of the version SCons has, is submodule able to track a single file from an external repo? Otherwise just main the version manually. Then the ninja tool can come packaged with SCons without the end-user needing to do any manual fetching or installing of the file.

Also I am wondering what the plan would be moving forward with maintenance of the ninja tool? 

1. Would MongoDB keep there own version, and SCons essentially pulls in updates from there, possibly diverging if SCons end users request other changes outside of MongoDB's version?
2. Would MongoDB switch to use the SCons tool instead of their site_tools version? It seems pretty essential to them, so could there be disagreements to the maintenance or direction of the tool in this scenario?

bdbaddog:
https://github.com/ninja-build/ninja is Apache licensed.
I'm not sure we can pull that code into the repo even with git submodules.
Thus the preference to add as a required package in setup.cfg.

I'd hope ongoing development would happen in the scons tree.
Thus far MongoDB's been taking the scons-local packages as we release new versions and patching whatever local changes they have.
I've been working with @acmorrow to upstream as much of they're patches and tools as is reasonable and that they'll support doing.

bdbaddog:
Moving to 4.1.0 Project. Unlikely this will be releasable in the very near term.
I'd like to get 4.0.0 out in the nearer term.

bdbaddog:
Closing this PR as @dmoody256  has a newer PR #3642  with it's content.

src/engine/SCons/Tool/ninja.py
Rdna123: File name update: [https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja.py](https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja.py)

bdbaddog: > File name update: https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja.py

This PR is closed and the work has been moved to PR #3642 please post your comment there and any further comments.

PR #3553, opened 2020-02-12
Do not reuse tempfiles

simontegelid:
Actions lists with more than one action using `${TEMPFILE("TEMP FILE CONTENT")}` should not use the same tempfile for different "TEMP FILE CONTENT".  Therefore, we cannot store the cmd_list in neither the target nor the source attributes, since those are common for all actions.

Reusing the same tempfile causes the same command to be invoked for all actions in the list. However, not resusing tempfiles will create unneeded tempfiles on `strfunction` invocation. This is better than producing incorrect results, but leaves the tempfile in a bad state.

Accidentally closed PR https://github.com/SCons/scons/pull/3140, but there is some relevant discussion in there.

## Contributor Checklist:

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -10,6 +10,9 @@ NOTE: Please include a reference to any Issues resolved by your changes in the b
 
 RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
+  From Simon Tegelid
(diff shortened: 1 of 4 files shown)


bdbaddog:
@simontegelid - I'm replacing this pull request with an alternate fix #3801 

Please take a look at that PR and maybe try on your build and let me know if you see any issues with the fix.

Issue #3626, opened 2020-04-23
Programmatic identification of conftest nodes


acmorrow:
**Describe the Feature**
Programmatic identification of conftest nodes.

Much like the issue reported in #3625, several of our tools work by interposing on emitters or scanners of existing builders. In some cases, the work we want to perform is inappropriate for conftests, either because of the special environment in which they execute, or because the injected behavior is inappropriate for conftests.

For now, we need to do something ugly like 

```
if not "conftest" in str(target[0):
    # Do stuff to non-conftest nodes
```

which is unpleasant and far from foolproof. We think it would be better if there were an attribute or method on Node that would allow us to identify conftest-related nodes directly. Ideally, we could write something better, like:

```
if not target[0].is_conftest():
    # Do stuff to non-conftest nodes
```

Alternatively:

```
if not getattr(target[0].attributes, "is_scons_conftest", False):
    # Do stuff to non-conftest nodes
```

Would it be possible to adorn all nodes used by the SConf subsystem this way? It would make it much easier to handle such nodes differently  when required, without needing to worry that someone is going to introduce a file called "conftest.cpp" next year when they want to write a test for some configuration, not knowing that that name has special meaning to SCons and our tools.

**Required information**
https://pairlist4.pair.net/pipermail/scons-users/2020-April/008072.html

Issue #3625, opened 2020-04-23
Programmatically expose all SConstruct/SConscript files used in the build.


acmorrow:
**Describe the Feature**

This request is driven by work on our Ninja builder for SCons. The Ninja builder needs to declare that the generated build.ninja file have a dependency on SConstruct and all the SConscripts, so that we can detect when the build.ninja file should be regenerated, because the structure of the build as described by SCons has changed.

Currently, we do this by globbing for such files, but this is not a great long term solution. In particular, env.SConscript allows the scripts to be named something other than SConscript, so if we want our Ninja tool to be truly generic, it shouldn't rely on filenames and globbing to do this work.

There are some tricky questions with this one, specifically, with how the phases of SCons execution would populate this vs when the Ninja backend would need access to the list.

Still, if there is still general interest in moving forward with getting the MongoDB Ninja generator upstreamed into SCons, some more robust solution to identifying the SCons files in play will be needed.

**Required information**
* Link to SCons Users thread discussing your issue.
https://pairlist4.pair.net/pipermail/scons-users/2020-April/008072.html

bdbaddog:
Does this include site_scons files as well?

acmorrow:
It is a great question and I can see arguments both ways. For SCons builds, any site_scons changes only cause a rebuild if they actually lead to a change in what/how things should be built. Ditto any changes to SConscripts or SConstruct. For Ninja, we just hard cause a regeneration if any of these files are out of date, but if the newly generated Ninja file doesn't change how any targets would be built, Ninja will consider them up to date. So I think that it would be preferable if site_scons loaded files were considered. This would, in effect, result in all of the built-in SCons tools being considered as well, right? So it would also probably capture a SCons upgrade as triggering a rebuild, which seems desirable.

bdbaddog:
Fixed in #3670

PR #3642, opened 2020-05-06
Ninja generation tool - experimental version

dmoody256:
**Summary**
This tool attempts to generically take a SCons build and generate a ninja file which is representative of the build and can be built by [Ninja](https://ninja-build.org/manual.html). 

Ninja is a very fast build system, which is meant to have its build files generated by other systems. Ninja is especially good at iterative rebuilds and quick traversing a build graph to determine what needs to be rebuilt.

This PR is to introduce the ninja tool based off the [ninja tool](https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/ninja_next.py) from mongodb's site tools.

This PR scope is **BETA functionality** of the ninja tool. It will not work well with all other tools and is primarily aimed at cross platform (win32/linux/osx) C/C++ builds. The goal of this PR is to get the tool to a point that it can be integrated in beta form.

**Feature/Test List**
Test Name | Test Description | Implemented
-- | -- | --
generate_and_build.py | Generate and build simple C program | X
generate_and_build_cxx.py | Generate and build simple CPP program and check header deps | X
generate_source.py | Generate source with built C program and build C program from generated source | X
multi_env.py | Build two C programs with two seperate envs with one ninja file | X
shell_commands.py | Simple shell command to take in sources and build targets | X
copy_function_command.py | SCons "Copy" function action converted to command line for ninja | X
build_libraries.py | Build and linking C shared and static libraries | X
iterative_speedup.py | Make sure Ninja is faster than SCons in single file change rebuild (Would love to see this fail one day) | X
output_orders.py | Make sure outputs are ordered correctly in ninja file |  
existing_targets.py | Make sure response files generate correctly even if existing targets are already built |  
unhandled_actions.py | Unhandled actions will reinvoke scons without ninja to build just those targets |  
ninja_rule.py | Create and register a custom ninja rule |  
command_line_targets.py | Specify Alias and targets from command to build |  
non_ninja_nodes.py | Make sure Scons is reinvoked on nodes which are not FS or Alias |  

**Known Issues**
Known Issue Description | Github Issue
-- | --
SCons doesn't have representive job pools to translate to ninja | #3667
Ninja tool is global, affects all environments regardless of what environment its invoked in |  
Compilation Databases interactions between SCons and Ninja |
Response files are always used even when not necessary |
Targets and -j are not propagated to ninja correctly |
TEMPLATE aggregation destroys order (need interactive mode) |

Contributor Checklist:

* [x] I have created a new test or updated the unit tests to cover the basic use.
* [ ] I have created a new test or updated the unit tests to cover ninja specific Methods.


dmoody256:
this test shows how fast ninja can be at iterative rebuilds, and it gets exponentially faster the more files involved: 
image

bdbaddog:
@dmoody256 - yup. not news. Ninja's raw speed is hard to beat.

dmoody256:
@bdbaddog yes, and that's why I'm excited SCons can leverage that speed! Ninja is meant to be fast and generated, and SCons has great syntax (python!), most extendable and broad feature support, now with this Ninja feature, it can crank through big builds as fast as possible.

bdbaddog:
Moving to 4.1.0 Project. Unlikely this will be releasable in the very near term.
I'd like to get 4.0.0 out in the nearer term.

Likely we'll push a beta with whatever of this is ready shortly after 4.0.0

dmoody256:
ok sounds good, I wrote some basic tests covering building basic C libs/applications. I still need to write some tests for the Ninja specific methods added to environment, and finishing documentation.

Overall the tool probably still needs some updates/tests regarding using other tools in conjunction with the ninja tool activated (example: other builders like Java, Fortran, etc..) since it reworks alot of SCons internals. That might not need to be handled on the PR, maybe just have a warning printed and documenting that this new feature is in a experimental phase.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+    # Make SCons node walk faster by preventing unnecessary work
+    env.Decider("timestamp-match")
+
+    # Used to determine if a build generates a source file. Ninja
+    # requires that all generated sources are added as order_only
+    # dependencies to any builds that *might* use them.
+    env["NINJA_GENERATED_SOURCE_SUFFIXES"] = [".h", ".hpp"]
+
+    if env["PLATFORM"] != "win32" and env.get("RANLIBCOM"):
```
acmorrow: This too belongs in a tool's customization of its interaction with the Ninja tool, and not in the Ninja tool itself.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+    # The "Alias Builder" isn't in the BUILDERS map so we have to
+    # modify it directly.
+    SCons.Environment.AliasBuilder.emitter = ninja_file_depends_on_all
+
+    for _, builder in env["BUILDERS"].items():
+        try:
+            emitter = builder.emitter
+            if emitter is not None:
+                builder.emitter = SCons.Builder.ListEmitter(
```
acmorrow: Using an emitter here seems not quite right. Should this instead be a scanner?

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+    # unnecessary work. If a build is a regular builder (i.e not a conftest and
+    # not our own Ninja builder) then we add it to the NINJA_STATE. Otherwise we
+    # build it like normal. This skips all of the caching work that this method
+    # would normally do since we aren't pulling any of these targets from the
+    # cache.
+    #
+    # In the future we may be able to use this to actually cache the build.ninja
+    # file once we have the upstream support for referencing SConscripts as File
+    # nodes.
+    def ninja_execute(self):
```
acmorrow: I think this could re-use more of SCons:

```
+
+    base_execute = SCons.Taskmaster.Task.execute
     def ninja_execute(self):
         global NINJA_STATE

         target = self.targets[0]
         target_name = str(target)
-        if target_name != ninja_file_name and "conftest" not in target_name:
-            NINJA_STATE.add_build(target)
-        else:
-            target.build()

+        if target_name == ninja_file_name or "conftest" in target_name:
+            base_execute(self)
+        else:
+            NINJA_STATE.add_build(target)

```

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+        else:
+            target.build()
+
+    SCons.Taskmaster.Task.execute = ninja_execute
+
+    # Make needs_execute always return true instead of determining out of
+    # date-ness.
+    SCons.Script.Main.BuildTask.needs_execute = lambda x: True
+
+    # We will eventually need to overwrite TempFileMunge to make it
```
acmorrow: This looks like an area that needs investigation regarding what we can do at the SCons layer to address it.

acmorrow:
This is a great start. Thanks @dmoody256 for putting it together and adding tests. There are a ton of things we need to investigate, improve, or re-design to make this a truly useful and general purpose tool. I've left a ton of comments trying to highlight these places and provide some context.

acmorrow:
Another thing to consider long term is whether the model and approach here could be extended to generate for other backends. Could we emit real msbuild or vcxproj this way?

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+        added = 1
+        AddOption('--disable-auto-ninja',
+                  dest='disable_auto_ninja',
+                  metavar='BOOL',
+                  action="store_true",
+                  default=False,
+                  help='Disable ninja automatically building after scons')
+    env["DISABLE_AUTO_NINJA"] = GetOption('disable_auto_ninja')
+
+    global NINJA_STATE
```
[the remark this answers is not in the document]
dmoody256: I don't really like it. One of the things I have been working on in my local is moving the NINJA_STATE into the specific environment that invoked the tool, so in theory in a given SCons build you can have separate environments in which one or more generates into a ninja file and the other environment does a normal scons build. Most other tools are pretty good about only affecting the environment they were invoked with. However there are a lot of ninja optimizations that change things for the Jobs, Taskmaster and other core SCons which are not easily decomposed into separate environments, so there would need to be some give and take there I think. I think for now I have tabled that idea for a later improvement.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+        target = [target]
+
+    # We have no COMSTR equivalent so print that we're generating
+    # here.
+    print("Generating:", str(target[0]))
+
+    generated_build_ninja = target[0].get_abspath()
+    NINJA_STATE.generate()
+
+    if env["PLATFORM"] == "win32":
```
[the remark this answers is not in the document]
dmoody256: As an end user, I found it a bit annoying that I had to always execute ninja after my scons command, so I put at as an extra step after scons is done generating the ninja file. I added a tool option to disable auto execution.

I also found that on windows, Ninja does not have a great way to take in an alternate env (Linux works with prepending environment variables to the command) so this would also run ninja in a subprocess with the correct environment as was setup with the scons build.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+def ninja_builder(env, target, source):
+    """Generate a build.ninja for source."""
+    if not isinstance(source, list):
+        source = [source]
+    if not isinstance(target, list):
+        target = [target]
+
+    # We have no COMSTR equivalent so print that we're generating
+    # here.
+    print("Generating:", str(target[0]))
```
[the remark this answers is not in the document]
dmoody256: this is called by the Ninja Builder so it should be possible to setup a COMSTR for that builders action

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+def exists(env):
+    """Enable if called."""
+
+    # This variable disables the tool when storing the SCons command in the
+    # generated ninja file to ensure that the ninja tool is not loaded when
+    # SCons should do actual work as a subprocess of a ninja build. The ninja
+    # tool is very invasive into the internals of SCons and so should never be
+    # enabled when SCons needs to build a target.
+    if env.get("__NINJA_NO", "0") == "1":
```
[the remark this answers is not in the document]
dmoody256: I wasn't sure exactly what the goal of this variable was, but seemed like it was to disable the ninja build if you invoked the build with ninja, which seemed an odd workflow?

dmoody256:
> Another thing to consider long term is whether the model and approach here could be extended to generate for other backends. Could we emit real msbuild or vcxproj this way?

SCons can already generate the MSVC projects, I'm not sure if by "real" your implying some discrepancy with the SCons generated MSVC projects.

I think generalizing an approach to generate different builds is a good path to take up. It would flush out issues in SCons by generalizing the build graph to mesh with other build systems and maybe give insights into techniques other build systems have leveraged.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+        
+    if NINJA_STATE is None:
+        NINJA_STATE = NinjaState(env, ninja_file[0], ninja_syntax.Writer)
+        NINJA_STATE.ninja_bin_path = os.path.abspath(os.path.join(ninja_syntax.__file__, os.pardir, 'data', 'bin'))
+    # Here we will force every builder to use an emitter which makes the ninja
+    # file depend on it's target. This forces the ninja file to the bottom of
+    # the DAG which is required so that we walk every target, and therefore add
+    # it to the global NINJA_STATE, before we try to write the ninja file.
+    def ninja_file_depends_on_all(target, source, env):
+        if not any("conftest" in str(t) for t in target):
```
acmorrow: Forgot to mention: this is where #3626 becomes important.

acmorrow:
> > Another thing to consider long term is whether the model and approach here could be extended to generate for other backends. Could we emit real msbuild or vcxproj this way?
> 
> SCons can already generate the MSVC projects, I'm not sure if by "real" your implying some discrepancy with the SCons generated MSVC projects.
> 

I think the SCons generated MSVC projects just re-invoke SCons though?

dmoody256:
> think the SCons generated MSVC projects just re-invoke SCons though?

ah yes, thats correct, SCons still issues the build commands.

SCons/Tool/ninja.py
```diff
@@ -1462,68 +1483,43 @@ def robust_rule_mapping(var, rule, tool):
...
+                ninja_bin))
+            if not os.path.exists(NINJA_STATE.ninja_bin_path):
+                # couldn't find it, just give the bin name and hope
+                # its in the path later
+                NINJA_STATE.ninja_bin_path = ninja_bin
+
+    # TODO: this is hacking into scons, preferable if there were a less intrusive way
+    # We will subvert the normal builder execute to make sure all the ninja file is dependent
+    # on all targets generated from any builders 
+    SCons_Builder_BuilderBase__execute = SCons.Builder.BuilderBase._execute
```
acmorrow: Why do it this way rather than just writing a scanner?

dmoody256: I am looking into the scanner method you suggested currently. I already had this implemented in my local and had been testing with it, so pushed it up before investigating the scanner route a bit more, but there are few draw backs with going through the env's list of builders.

One problem with just using the list of builders in the current environment, is it only captures the builders at the time the ninja tool was generated, so in a given SCons build, if you add builders later, they don't get emitters/scanners updated to make everything depend on the ninja file. Also the ninja tool affects SCons not just in the environment it was generated with, it affects all environments, so another reason going through the passed in environment's builder list is not reliable.

Secondly, there are other builders that will never be in the env's builder list, such as the Alias builder or the Command builder, so alternate solutions need to be done for those. There may be other builders too that I am not aware of that are never added to the builder list.

This solution, every builder will use this execute function, and this happens after the scons scripts have been read and executed so you can add builders after the ninja tool is initialized.

Ideally the instead of forcing this change directly into the SCons internal class from the tool, SCons has some new function that allows some callback to be setup, or lets you set all emitters/scanners for all builders.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+        elif results[0]["rule"] == "INSTALL":
+            return {
+                "outputs": all_outputs,
+                "rule": "INSTALL",
+                "inputs": [get_path(src_file(s)) for s in node.sources],
+                "implicit": dependencies,
+            }
+
+        raise Exception("Unhandled list action with rule: " + results[0]["rule"])
```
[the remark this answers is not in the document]
dmoody256: Why does the handle_list_action function only check the rule of the first result for its list of results? It calls the action_to_ninja_build function for each item which may return a different ninja rule for each item in results. 

I am understanding this function to 

1. take in a list action and convert each action in the list to a ninja build dictionary
2. don't actually use the series of dictionary in those results, but generate a new single ninja build dictionary to represent all of them

So if we only check the first item we, are making some assumptions that the rest are of that same type?

Could instead the several ninja build dictionaries be made to depend on each other so they execute in the same order? Maybe some of the current handling are optimization cases where its better to have a single build item, but those could still be handled separately, and instead of just raising exception at the end here we generate all the ninja build items and make them depend on each other?

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+from os.path import join as joinpath
+from os.path import splitext
+
+import SCons
+from SCons.Action import _string_from_cmd_list, get_default_ENV
+from SCons.Util import is_List, flatten_sequence
+from SCons.Script import COMMAND_LINE_TARGETS
+
+NINJA_STATE = None
+NINJA_SYNTAX = "NINJA_SYNTAX"
```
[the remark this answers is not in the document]
dmoody256: as of bbec480, the generate function is trying to import ninja and grab the syntax module from there. NINJA_SYNTAX is not used anymore, and I will update to remove it completely.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+    outputs = [link.get_dir().rel_path(linktgt) for link, linktgt in symlinks]
+    inputs = [link.get_path() for link, _ in symlinks]
+
+    return {
+        "outputs": outputs,
+        "inputs": inputs,
+        "rule": "SYMLINK",
+        "implicit": get_dependencies(node),
+    }
```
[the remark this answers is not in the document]
dmoody256: If alot of the SCons function actions would be converted to command actions, I wonder the level of negative impact to performance for normal SCons builds if they need to spawn new shells more often? Would anyone or anything else prefer the command actions over function actions besides the ninja tool?

If its significant impact, maybe function actions could have command action variants? However, I don't like the idea of maintaining the same feature in two different implementations.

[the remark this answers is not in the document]
bdbaddog: @dmoody256 - I did that with Install/InstallAs to check perf. And the intrinsic python logic was the same or slightly faster. (Surprisingly) even on linux.. Even for parallel builds.

Realistically, there will be almost no maintenance. What's going to change with logic to copy a file? As long as the action is well defined, a note in the sources specifying that the logic in Ninja generator has to be kept in sync if the functionality changes should suffice.

Or.. An equivalent command line action could be annotated to the python action which the ninja (or other) generator can use.  At least that way the logic is all in one place
(Think python decorator).

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+            #
+            # (The deps section)
+            "msvc_deps_prefix": "Note: including file:",
+        }
+
+        self.rules = {
+            "CMD": {
+                "command": "cmd /c $env$cmd $in $out" if sys.platform == "win32" else "$env$cmd $in $out",
+                "description": "Building $out",
+                "pool": "local_pool",
```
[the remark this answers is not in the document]
dmoody256: Wrote an issue here to add pools to SCons: #3667

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+                "rspfile_content": "$rspc",
+                "pool": "local_pool",
+            },
+            "AR": {
+                "command": "$env$AR @$out.rsp",
+                "description": "Archiving $out",
+                "rspfile": "$out.rsp",
+                "rspfile_content": "$rspc",
+                "pool": "local_pool",
+            },
```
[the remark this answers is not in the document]
dmoody256: The way I imagine it working is the ninja tool queries the list of tools from the environment, and then try to call some common function like get_ninja_rules on each tool, and just passes on exception. Each tool can then add there own rules to the ninja tool, and maintain the rules separately. 

There would probably need to be special handling when a tools define a similar rule, although this already happens with some of the env variables that tools overlap on, which just do an order dependent overwrite in most cases.

That idea seems dependent on each environment being able to generate its own ninja file?

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+                # Again we restat in case Ninja thought the
+                # build.ninja should be regenerated but SCons knew
+                # better.
+                "restat": 1,
+            },
+        }
+
+        self.pools = {
+            "local_pool": self.env.GetOption("num_jobs"),
+            "install_pool": self.env.GetOption("num_jobs") / 2,
```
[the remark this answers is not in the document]
dmoody256: Wrote an issue for scons pools: #3667

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+
+    return wrapper
+
+
+def ninja_contents(original):
+    """Return a dummy content without doing IO"""
+
+    def wrapper(self):
+        name = str(self)
+        if "SConscript" in name or "SConstruct" in name:
```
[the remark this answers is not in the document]
dmoody256: This is related to making it possible to reliably detect Nodes to be SConstruct or SConscript types from issue #3625. That issue doesn't seem terribly difficult without really digging into it. I think we should implement that before this ninja tool beta version is merged.

acmorrow:
@dmoody256 - FYI, please be advised of the following fix that landed in the MongoDB version of this: https://github.com/mongodb/mongo/commit/18cbf0d581162b2d15d66577b1fe08fe22006699#diff-e5450d4ac791a19fac9a32fac3a1b19c

Another one incoming that looks like this:

diff --git a/site_scons/site_tools/ninja_next.py b/site_scons/site_tools/ninja_next.py
index f999b746560..0ce56407439 100644
--- a/site_scons/site_tools/ninja_next.py
+++ b/site_scons/site_tools/ninja_next.py
@@ -1358,6 +1358,16 @@ def generate(env):
     SCons.Node.FS.Dir.get_csig = ninja_csig(SCons.Node.FS.Dir.get_csig)
     SCons.Node.Alias.Alias.get_csig = ninja_csig(SCons.Node.Alias.Alias.get_csig)

```
+    # Ignore CHANGED_SOURCES and CHANGED_TARGETS. We don't want those
+    # to have effect in a generation pass because the generator
+    # shouldn't generate differently depending on the current local
+    # state. Without this, when generating on Windows, if you already
+    # had a foo.obj, you would omit foo.cpp from the response file. Do the same for UNCHANGED.
+    SCons.Executor.Executor._get_changed_sources = SCons.Executor.Executor._get_sources
+    SCons.Executor.Executor._get_changed_targets = SCons.Executor.Executor._get_targets
+    SCons.Executor.Executor._get_unchanged_sources = SCons.Executor.Executor._get_sources
+    SCons.Executor.Executor._get_unchanged_targets = SCons.Executor.Executor._get_targets
+
     # Replace false action messages with nothing.
     env["PRINT_CMD_LINE_FUNC"] = ninja_noop
```

dmoody256:
@acmorrow thanks! 

I would like to add some tests for these cases to capture this functionality.  For the sorted issue, from my understanding it caused the generated ninja file to put the outputs in the wrong order to some build command that was expecting dwo to come after the .o? I would like to mock it up with some script that expects args in a certain order, but want to make sure that was the issue.

acmorrow:
@dmoody256 - No, not exactly. It was more that having the `.dwo` file as the "real" target (as opposed to using the `phony` mechanism we use for mulitple outputs` appeared to completely disable ninja deps generation. The result was that ninja would not see targets as out of date when the should have been.

acmorrow:
Also @dmoody256, this commit did land in the MongoDB version: https://github.com/mongodb/mongo/commit/a7541c60e5904e7deafdedf5bb040cc8924ac7d3

SCons/Tool/ninja.py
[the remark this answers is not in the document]
acmorrow: I think you should consider not doing this:

- Yes, you need to re-invoke Ninja. But that is the right thing to do. You can't know what targets the user actually wants to build. The only target you can name to SCons when building is the ninja file itself, so you have no information about what they actually want to build. The default target might be expensive to build, and maybe they just wanted a particular unit test.

- Regarding the windows environment - yeah, that is a thing. But if you wanted a hermetic build, you should have stayed with SCons driving the build. Once you engage with Ninja, you are building a snapshot and you are at the mercy of the local environment. We actually do manage to pass environment variables down in the generated Ninja on Windows, btw.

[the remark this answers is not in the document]
dmoody256: > The only target you can name to SCons when building is the ninja file itself, so you have no information about what they actually want to build.

I found this a bit confusing that you had to name ninja-generation as a target to generate ninja for the mongodb build. I agree that the ninja file is an output of the build, but I was approaching the ninja file not like a real target. In my opinion its more like an extra build step to get to the actual output of the build. I don't see why targets provided to SCons could not be passed down to ninja command? SCons should also be able to tell if the ninja file needs to be regenerated and just skip to ninja execution with the new targets if not. Anyways, this was more of a personal UIX thing, so its not really functional. I am completely fine with not doing it. 

> But if you wanted a hermetic build, you should have stayed with SCons driving the build.

What if I want a hermetic build and ninja speed? I don't think its impossible, maybe it should be deferred and voted on, but I don't think that's unreasonable want.

> We actually do manage to pass environment variables down in the generated Ninja on Windows, btw.

Any tips on where and what that looks like? I am trying to setup the tool so that a new user to SCons can almost effortless get a SCons with ninja build working. Needing to preconfigure a build environment is not a pro, but if there is a way for the build to do this for the user that is a pro.

fire:
To be clear this handles the two use cases?

1. I'm able to open the ninja - (ninja has an option to spit out a compilation db) file into clion and it can build the entire project with code completion.
1. The ninja compilation is able to be distributed via icecream.

dmoody256:
@fire 
> I'm able to open the ninja - (ninja has an option to spit out a compilation db) file into clion and it can build the entire project with code completion.

I am not super familiar with clion, so I can't say for sure, but the tool does tell ninja to generate the compilation database. It uses response files heavily so hopefully clion doesn't have issue with that.

> The ninja compilation is able to be distributed via icecream.

The tool itself may not have knowledge of other distributed compilers, but it should be customize-able through the environment variables specifically by changing the CCOM or CXXCOMs. Mongodb has done this with their local tool and I believe it works with their local ninja tool: https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/icecream.py#L260-L263

fire:
The title is beta, does this mean I can test this? Can you write some instructions?

dmoody256:
> The title is beta, does this mean I can test this? Can you write some instructions?

I haven't written documentation yet, but for some examples you can check out the test examples. Here I copied and pasted a super simple example, basically just initialize the ninja tool in your environment and build:
```
env = Environment()
env.Tool('ninja')
env.Program(target = 'foo', source = 'foo.c')
```

**Note:** You need to install ninja with pip: ```pip install ninja``` and make sure that the python you installed ninja to is running scons. Currently initializing Ninja is global within scons, so all environments will be affected and their targets will be generated to the ninja file. Also any custom function actions or list actions may not generate to the ninja file, a handler needs to be written to process the function action into the command line version for ninja to execute.

fire:
I ran into some problems with Godot Engine and scons ninja:

* Disabled existing compilation db support
* `Found unhandled function action`, couldn't find a way to pass through scons functions used to generate c++ files. Is it possible to call scons for the function? It's not ideal but it would break existing code less

acmorrow:
@fire - The ninja generator _should_ be invoking SCons again for any unhandled function actions. But that is slower than having Ninja do it. If you convert your function action into a command generator, then it is no longer opaque to the Ninja Generator and it can probably handle it. If the Ninja Generator isn't handling the fallback, that would be a bug.

fire:
It calls an exception here. https://github.com/dmoody256/scons/blob/ninja-generation/SCons/Tool/ninja.py#L283

acmorrow:
@fire - Ah, I see. I guess we decided to make that a hard error rather than silently generate an inferior build, which sort of makes sense. Anyway, your two options then are to:
- Rewrite your function actions to be command generator actions, or
- Do as the exception suggests, and use `NinjaRegisterFunctionHandler` to register your own handler for your function action, which should emit the necessary Ninja declarations to do work equivalent to what your function action would have done. You can find some example usages in the MongoDB repo (https://github.com/mongodb/mongo/).

fire:
I got stuck writing my own handlers, so if you could provide one that can handle any call even though inferior, I can piecemeal convert. 

Currently, I can't demonstrate that ninja works when the generated c++ code causes the build to fail.

acmorrow:
I think it is going to look something like this, but I haven't tested it, and I won't have time to right away. All this stuff is super new and still under active development. It will be beta in the next release of SCons. It is somewhere around alpha now.

Anyway, if you have `my_function_action` that you want to redirect to SCons, I think you can do it like follows:

```
    def invoke_scons_for_function_action(env, node):
        """Generates empty .a files"""
        return {
            "outputs": [node.get_path()],
            "rule": "SCONS",
            "implicit": [str(s) for s in node.sources],
        }

    env.NinjaRegisterFunctionHandler("my_function_action", invoke_scons_for_function_action)
```

Now I think this will result in a re-invocation of SCons for every one of those targets, which is sort of brutal. It may be that you can consolidate them by saying `TEMPLATE` there instead of `SCONS` for the `rule`.

dmoody256:
@fire 
> Currently, I can't demonstrate that ninja works when the generated c++ code causes the build to fail.

godot was one of the big scons builds I knew about that I was going to test this tool against before merging, so its great that someone more familiar with godot is partaking. Thanks for taking some time.

@acmorrow 
> Now I think this will result in a re-invocation of SCons for every one of those targets, which is sort of brutal.

Maybe some patch could be made to SCons to supply a arg, which doesn't do the full dependency walk for a target to determine dependencies, and instead just executes single target assuming the sources are in place and built? It might be more trouble then its worth, but I might take a look down that path.

acmorrow:
> @fire
> 
> > Currently, I can't demonstrate that ninja works when the generated c++ code causes the build to fail.
> 
> godot was one of the big scons builds I knew about that I was going to test this tool against before merging, so its great that someone more familiar with godot is partaking. Thanks for taking some time.
> 
> @acmorrow
> 
> > Now I think this will result in a re-invocation of SCons for every one of those targets, which is sort of brutal.
> 
> Maybe some patch could be made to SCons to supply a arg, which doesn't do the full dependency walk for a target to determine dependencies, and instead just executes single target assuming the sources are in place and built? It might be more trouble then its worth, but I might take a look down that path.

I don't think that is necessary. The Ninja tool has logic to gather the `TEMPLATE` builders into one SCons invocation. I think `TEMPLATE` here is intended to mean "generated". I think that is the right thing. Part of the difficulty is that the original author of this tool is no longer involved, so I'm having to reverse engineer some pieces of it with which I wasn't fully familiar. The details around handing and interposing on function actions, and how and when we re-invoke SCons are aspects that I need to brush up on.

fire:
Here's changes to godot 3.2 stable if you're curious.

https://github.com/godot-extended-libraries/godot-fire/commit/202dea34e3da9410eaf3925fcade3e2cec25a4ad

dmoody256:
@fire 
I got your changes building with ninja with a slightly hacked version of the ninja tool here, it includes some newer stuff as well than what is in this PR currently: https://github.com/dmoody256/scons/tree/ninja-tool-godot

I turned on reinvokation of SCons for the non handled function actions, and then also ran into an issue where it seems your build is using Python.Value nodes, which the ninja tool was not equipped to handle. I just skipped over those nodes when determining dependencies for ninja for now, so the build may not work completely correctly until I figure out a way to handle those nodes.

Actually currently that branch is always rebuilding a good portion of the godot build, which I suspect is from the reinvokation or the missing Python.Value nodes. I'll take a deeper look when I get more time.

fire:
TL;DR Better, but not there yet.

1. Used https://github.com/godot-extended-libraries/godot-fire/commit/202dea34e3da9410eaf3925fcade3e2cec25a4ad
2. Use https://github.com/dmoody256/scons/tree/ninja-tool-godot
3. pip install -e scons
3. Opened windows native tools prompt
4. scons
5. modified build.ninja
```
SCONS_INVOCATION = "C:\Users\fire\scoop\apps\python\current\Scripts\scons" "target=release_debug" $
    --disable-ninja __NINJA_NO=1 $out
SCONS_INVOCATION_W_TARGETS = "C:\Users\fire\scoop\apps\python\current\Scripts\scons.exe" "target=release_debug" --disable-ninja
msvc_deps_prefix = Note: including file:
```
6. ninja

```
Generating Code...
Compiling...
microsoft
c1xx : fatal error C1083: Cannot open source file: 'files\microsoft': No such file or directory
visual
c1xx : fatal error C1083: Cannot open source file: 'visual': No such file or directory
studio
c1xx : fatal error C1083: Cannot open source file: 'studio': No such file or directory
include
c1xx : fatal error C1083: Cannot open source file: '10.0\VC\include': No such file or directory
Microsoft
c1xx : fatal error C1083: Cannot open source file: 'files\Microsoft': No such file or directory
include
c1xx : fatal error C1083: Cannot open source file: 'SDKs\Windows\v7.0A\include': No such file or directory
Files
c1xx : fatal error C1083: Cannot open source file: 'Files': No such file or directory
Windows
c1xx : fatal error C1083: Cannot open source file: '(x86)\Windows': No such file or directory
include
c1xx : fatal error C1083: Cannot open source file: 'Kits\10\include': No such file or directory
Generating Code...
ninja: build stopped: subcommand failed.
```

dmoody256:
@fire 
I was testing linux, and those look like pathing issues, i'll try it out on windows.

bdbaddog:
@fire - Can you pastebin your ninja file somewhere and post a link?

fire:
This isn't what you asked for, but as I'm currently on the mac, I'll post it.

[ninja.log](https://github.com/SCons/scons/files/4720835/ninja.log)
[scons.log](https://github.com/SCons/scons/files/4720836/scons.log)
[build.ninja.log](https://github.com/SCons/scons/files/4720838/build.ninja.log) Rename to build.ninja.

Will be on the pc later in the week.

fire:
Here's the win10 log.

[build.ninja.log](https://github.com/SCons/scons/files/4720966/build.ninja.log)

acmorrow:
Hi @dmoody256 - Could you make sure that these changes from our repo get folded in? They fix important bugs:
- https://github.com/mongodb/mongo/commit/18cbf0d581162b2d15d66577b1fe08fe22006699#diff-e5450d4ac791a19fac9a32fac3a1b19c
- https://github.com/mongodb/mongo/commit/a7541c60e5904e7deafdedf5bb040cc8924ac7d3#diff-e5450d4ac791a19fac9a32fac3a1b19c

That second one especially looks like it could have something to do with @fire's problem above.

dmoody256:
@acmorrow 

Those commits where in my edge branch https://github.com/dmoody256/scons/tree/ninja-tool-godot (along with some other fixes I was testing).

I went ahead and pulled in the commits you had recommended to this PR.

fire:
@dmoody256 If you need anything else, feel free to ask. It seems to be related to 1. python not able to execute a binary executable on windows. 2. paths are not quoted.

dmoody256:
@fire 
I pushed up some path escaping for windows on that edge branch: https://github.com/dmoody256/scons/commit/0333b04fe0de9bbfed4ff3a2e0b0d04d0dd23d86

Still the build is not working completely correctly, because it is always rebuilding a 3rd of the build. Still suspect are reinvokation or Python.Value nodes. When I have some more time I will look into that.

fire:
1. git clone https://github.com/dmoody256/scons -b ninja-tool-godot 
1. pip install -e scons
1. Clone https://github.com/godot-extended-libraries/godot-fire/commit/202dea34e3da9410eaf3925fcade3e2cec25a4ad
1. open native tools 64 bit msvc
1. scons 
1. edit build.ninja to remove "c:\users\fire\scoop\apps\python\3.8.3\python.exe" $
    "C:\Users\fire\scoop\apps\python\current\Scripts\scons"
1. Should be "C:\Users\fire\scoop\apps\python\current\Scripts\scons" only.
1. ninja
1. Run bin\godot.windows.tools.64.exe

Appears to work!
```
powershell -Command "Measure-Command {ninja}"
<snipped>
TotalSeconds      : 119.7736507

>bin\godot.windows.tools.64.exe
Godot Engine v3.2.2.beta.custom_build.202dea34e - https://godotengine.org
```

fire:
What are Python.Value nodes? How are they used in scons?

bdbaddog:
@fire please go read the docs. 
https://scons.org/doc/production/HTML/scons-man.html#f-Value

dmoody256:
@fire 
the godot-fire repo you linked is using the mongodb compilation_db site_tool which is what uses the Python.Value nodes, but mongodb disables compilation_db site_tool when doing ninja build in favor of ninja's compilation db feature. It would then make sense that godot build would disable it in the ninja build case also. 

Edit:
Actually @acmorrow discussed this a bit in his review, and there are still a few remaining design questions: https://github.com/SCons/scons/pull/3642#discussion_r424542869

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+                self.env.File("#SConstruct").path,
+                __file__,
+            ]
+            + sorted(glob("src/**/SConscript", recursive=True)),
+        )
+
+        # If we ever change the name/s of the rules that include
+        # compile commands (i.e. something like CC) we will need to
+        # update this build to reflect that complete list.
+        ninja.build(
```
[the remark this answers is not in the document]
fire: Which method is the most accurate? Currently the ninja db hides the files in response files so it's more difficult to unpack.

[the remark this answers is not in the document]
dmoody256: For the expansion in compilation database, Ninja has the -x option for compdb added in this PR (I didn't see where that was actually documented): https://github.com/ninja-build/ninja/pull/1223

I think both will need to be enabled because for unhandled actions, Ninja falls back to SCons to build so for some builds, the Ninja comiledb by itself is not the complete build. I think only the ninja tool needs to detect and react if the compilation_db tool is in the environment, but detection is not straight forward (depends on the order the tools are initialized) and in the current BETA state, ninja tool affects all environment while the SCons compilation db tool may not.

The ninja tool could potentially then perform a final combination of the compilation data from both the SCons and Ninja compiledb output.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+                ),
+            ),
+            "SCONS_INVOCATION_W_TARGETS": "{} {}".format(
+                sys.executable, " ".join([escape(arg) for arg in sys.argv])
+            ),
+            # This must be set to a global default per:
+            # https://ninja-build.org/manual.html
+            #
+            # (The deps section)
+            "msvc_deps_prefix": "Note: including file:",
```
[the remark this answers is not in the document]
dmoody256: added default and environment get in 51a430b3abb0a0d4be5cfcf5a8250971e80a8727

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+            # before trying to check if it has a deps key.
+            #
+            # Anything using deps or rspfile in Ninja can only have a single
+            # output, but we may have a build which actually produces
+            # multiple outputs which other targets can depend on. Here we
+            # slice up the outputs so we have a single output which we will
+            # use for the "real" builder and multiple phony targets that
+            # match the file names of the remaining outputs. This way any
+            # build can depend on any output from any build.
+            build["outputs"].sort()
```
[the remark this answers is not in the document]
dmoody256: pulled in fix from mongodb: a0acd13fa03b4984938c524695f7cb5f46256dcd

fire:
I've been trying Scons with Ninja for about two weeks.

The only significant issue seems to be rebuilds don't appear to be optimal.

There was mention that the build has room for a 1/3 improvement, but there is anecdotal conversations that suggest ninja makes the Godot Engine build faster.

There seems to be checks that are failing in CICD.

Is it possible to include the ninja tool as an option in Scons instead of via the tool include?

bdbaddog:
@fire what do you mean by "Is it possible to include the ninja tool as an option in Scons instead of via the tool include?"

fire:
I need to add `env_base.Tool('ninja')` to the Godot Engine build.

Can we support taking any scons project without modification and being able to use ninja?

bdbaddog:
@fire - no it will not likely ever be a default builder as it adds hooks into many parts of scons which are only needed for ninja.

fire:
Why is that? I know that other build systems allow you to choose between the default on the platform and multiple platforms like json_compilation.db and ninja.

From my trial in Godot Engine, the `https://github.com/dmoody256/scons -b ninja-tool-godot` only required adding the `env_base.Tool('ninja')` line and stripping Godot's incompatible json_compilation.json work.

What do you mean by  `adding hooks into many parts of scons which are only needed for ninja.`?

There can be a flag that toggles ninja on but leaves the default scons builder as default. The concept is that something like ```scons --generator ninja``` will inject `env.Tool('ninja')` into the scons build.

bdbaddog:
@fire - Yes. You'll have to add env.Tool('ninja')
It won't be a tool that is in the default list to be always available without an explicit Tool() call.

You can refer to the source code for the ninja tool for details on how it's implemented if you really want to understand.

We're not other build systems.  
So saying other build systems can do A why can't yours isn't really helpful.
Indeed SCons does things that other build systems don't and can't do. (again.. I'm not going to list these, but you're free to read the sources and the documentation)

We can do what we can do with the available developers and no more. (This is true of any open source project).

There's no reason you can't add ```--generator`` on your own.

The ninja generator will be BETA in SCons 4.1.0.
We're still working on SCons 4.0.0
It will be non-beta most likely in 4.2.0.
(Those releases have no fixed dates so don't ask for them.. see statement about available developers above)

You're feedback on this PR is most welcome however.
Ensuring that it is useful for as many projects and usages as possible is why PR's exist.. Especially [WIP] ones..

setup.cfg
```diff
@@ -41,7 +41,9 @@ classifiers =
 [options]
 zip_safe = False
 python_requires = >=3.5
-install_requires = setuptools
+install_requires = 
+    setuptools
+    ninja
```
bdbaddog: We shouldn't require ninja for all scons installs. Please remove.

bdbaddog: The tool itself should check for availability like all such tools in it's exists().

bdbaddog:
@fire - maybe after Ninja is in the main release and not a beta feature would --generator make sense. But certainly not at this point.

setup.cfg
dmoody256: removed and added check in exists: 
c10303d
2419d9c

bdbaddog:
A couple of macro comments.

1. ninja.py is the 2nd largest single file tool (in ```SCons/Tools```), I'd like to see the bulk of the functionality moved into SCons/Tools/ninja/... hopefully with just a simplified exists and generate in the actual ninja.py
2. Minor one.. in all output ideally scons is output as SCons
3. Would be nice to have docstrings on all functions. This is complicated enough logic that having it show up in API docs might be worthwhile.

mwichmann:
>     3. Would be nice to have docstrings on all functions. This is complicated enough logic that having it show up in API docs might be worthwhile.

Just note that the new Sphinx build excludes any Tools from building into the API docs, just like the old Epydoc build did, so they still wouldn't show up there...

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+        self.built = set()
+
+        # SCons sets this variable to a function which knows how to do
+        # shell quoting on whatever platform it's run on. Here we use it
+        # to make the SCONS_INVOCATION variable properly quoted for things
+        # like CCFLAGS
+        escape = env.get("ESCAPE", lambda x: x)
+
+        self.variables = {
+            "COPY": "cmd.exe /c 1>NUL copy" if sys.platform == "win32" else "cp",
```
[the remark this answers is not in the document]
bdbaddog: yes should respect SPAWN/SHELL

bdbaddog:
Can you add ninja to the requirements.txt file? (With a comment that it's only needed for ninja tool)
Also we're not **yet** adding it to the setup.cfg to force install when SCons is installed.

fire:
I'm wondering what steps is blocking merging?

bdbaddog:
@fire - my time.
Also there's nothing preventing you from using this code as is today.

SCons/Tool/ninja.py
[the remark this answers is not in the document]
chasinglogic: > Could instead the several ninja build dictionaries be made to depend on each other so they execute in the same order? 

This is how it used to work and it was too hard to get a ninja file that both functioned properly (had correct DAG) and that Ninja would happily parse. In the interest of shipping I changed this to the simpler solution here which doesn't break on the Mongo code base (which was our smaller subset of caring at the time).

Also a trick is that some things are `ListActions` that don't need to be, for instance Linking on Windows is a `ListAction` where the second Action is a `FunctionAction` which we had no need or analogy for. ([here are the relevant functions](https://github.com/SCons/scons/blob/master/SCons/Tool/mslink.py#L215))

tl;dr It can be done that way but it was more fiddly than I had time for when originally writing this tool.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+            return False
+
+        node_string = str(node)
+        if node_string in self.builds:
+            raise Exception("Node {} added to ninja build state more than once".format(node_string))
+        self.builds[node_string] = build
+        self.built.update(build["outputs"])
+        return True
+
+    def is_generated_source(self, output):
```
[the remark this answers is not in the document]
chasinglogic: > the challenge is that there is no good way to describe at the Ninja layer interdependencies between generated files

Yes Ninja has absolutely no idea how to handle generated files and so this is how every ninja generator handles this case.

> The current detection for what counts as a generated source is weak.

Agreed 🤷

SCons/Tool/ninja.py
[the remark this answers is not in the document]
chasinglogic: It's a way for ninja to run SCons and indicate to the ninja generator it should not disable building.

The way the Ninja generator works is it prevents SCons builders from actually executing (importantly this includes FunctionActions). For FunctionActions / Subst builders the generated ninja file will re-invoke SCons to build those targets. However because we carte blanche copy the SCons flags and vars the generator was invoked with this would mean the generator would still invade and disable this behaviour. So if you had a target `some-substed-file.txt` then ninja will invoke SCons to build that file however if the ninja generator is on `some-substed-file.txt` would never be built.

So we inject this in the generated SCons commands in the ninja file so the ninja generator knows that it's being invoked as a sub-process of ninja and doesn't disable building.

SCons/Tool/ninja.py
```diff
@@ -0,0 +1,1570 @@
...
+    # accidental deletion of the CC/XXCOM mappings. You can still
+    # overwrite them if you really want to but you have to explicit
+    # about it this way. The reason is that if they were accidentally
+    # deleted you would get a very subtly incorrect Ninja file and
+    # might not catch it.
+    env.AddMethod(register_custom_rule_mapping, "NinjaRuleMapping")
+
+    # TODO: change LINKCOM and SHLINKCOM to handle embedding manifest exe checks
+    # without relying on the SCons hacks that SCons uses by default.
+    if env["PLATFORM"] == "win32":
```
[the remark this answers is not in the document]
chasinglogic: It is see my comment on the ListActions section above

bdbaddog:
> > Another thing to consider long term is whether the model and approach here could be extended to generate for other backends. Could we emit real msbuild or vcxproj this way?
> 
> SCons can already generate the MSVC projects, I'm not sure if by "real" your implying some discrepancy with the SCons generated MSVC projects.
> 
> I think generalizing an approach to generate different builds is a good path to take up. It would flush out issues in SCons by generalizing the build graph to mesh with other build systems and maybe give insights into techniques other build systems have leveraged.

SCons doesn't create "real" MSVC projects. It just wraps calling SCons.   You could just as easily wrap calling SCons in Ninja where the rule would just call SCons.  So what I'm talking about is more generalization of the build steps mapping to some file format. 
Of course this is beyond the scope of the PR. But just something to think about for the future..

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,293 @@
...
+                in this case the default target name will be used.
+            </para>
+            <note>
+                <para>
+                    You must load the &t-ninja; tool prior to specifying
+                    any part of your build or some source/output
+                    files will not show up in the compilation database.
+                </para>
+                <para>
+                    To use this tool you must install pypi's <ulink url="https://pypi.org/project/ninja/">ninja
```
mwichmann: probably should get a manpage mention, since early on that doc says should be no other dependencies.

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,293 @@
...
+            <item>NINJA_SYNTAX</item>
+
+            <item>_NINJA_REGENERATE_DEPS_FUNC</item>
+        </sets>
+
+        <uses>
+            <item>CCCOM</item>
+            <item>CXXCOM</item>
+            <item>SHCXXCOM</item>
+            <item>CC</item>
```
mwichmann, line 90: not SHCCCOM?

doc/user/external.xml
```diff
@@ -283,4 +285,24 @@ env2.CompilationDatabase('compile_commands-linux64.json')
...
   </section>
+
+    <section>
+        <title>Ninja Build Generator</title>
+
+        <para>
+            This is an experimental new feature.
+        </para>
+        <para>
+            Using the
```
mwichmann: placeholder content? Seems odd to have four such short paragraphs...

SCons/Tool/ninja/NinjaState.py
```diff
@@ -0,0 +1,708 @@
...
+import io
+import os
+import sys
+from os.path import splitext
+import ninja
+
+import SCons
+from SCons.Script import COMMAND_LINE_TARGETS
+from SCons.Util import is_List
+import SCons.Tool.ninja.Globals
```
mwichmann: line needed?

SCons/Tool/ninja/NinjaState.py
```diff
@@ -0,0 +1,708 @@
...
+from .Globals import COMMAND_TYPES, NINJA_RULES, NINJA_POOLS, \
+    NINJA_CUSTOM_HANDLERS
+from .Rules import _install_action_function, _mkdir_action_function, _lib_symlink_action_function, _copy_action_function
+from .Util import get_path, alias_to_ninja_build, generate_depfile, ninja_noop, get_command, get_order_only, \
```
mwichmann: since this is new code, may as well use Black-style setup for importing many symbols, then some future reformat won't cause a change:
```
from .Rules import (
    _install_action_function,
    _mkdir_action_function,
    _lib_symlink_action_function,
    _copy_action_function,
)
from .Util import (
    get_path,
    alias_to_ninja_build,
    generate_depfile,
    ninja_noop,
    get_command,
    get_order_only,
    get_outputs,
    get_inputs,
    get_dependencies,
    get_rule,
    get_command_env,
)
```

SCons/Tool/ninja/NinjaState.py
```diff
@@ -0,0 +1,708 @@
...
+
+        elif results[0]["rule"] == "INSTALL":
+            return {
+                "outputs": all_outputs,
+                "rule": get_rule(node, "INSTALL"),
+                "inputs": get_inputs(node),
+                "implicit": dependencies,
+            }
+
+        raise Exception("Unhandled list action with rule: " + results[0]["rule"])
```
mwichmann, line 711: maybe InternalError? Or ise it a UserError?

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,293 @@
...
+                If called with multiple sources,
+                the source list will be ignored,
+                since there is no way to deduce what the intent was;
+                in this case the default target name will be used.
+            </para>
+            <note>
+                <para>
+                    You must load the &t-ninja; tool prior to specifying
+                    any part of your build or some source/output
+                    files will not show up in the compilation database.
```
mwichmann: is this an unedited copy-paste from the compilationdb page?

bdbaddog: Docs are not ready for consumption mostly. So it could be.

SCons/Tool/ninja/ninja.xml
bdbaddog: This will be in the next push.

doc/user/external.xml
bdbaddog: I've filled this in a bit in the next push.

SCons/Tool/ninja/NinjaState.py
bdbaddog: Fixed in next push.

SCons/Tool/ninja/NinjaState.py
bdbaddog: Fixed in next push.

SCons/Tool/ninja/ninja.xml
bdbaddog: Fixed in next push.

SCons/Script/SConsOptions.py
```diff
@@ -39,7 +39,7 @@
 
 diskcheck_all = SCons.Node.FS.diskcheck_types()
 
-experimental_features = {'warp_speed', 'transporter'}
+experimental_features = {'warp_speed', 'transporter', 'ninja'}
```
mwichmann, line 42: are you intending to leave the two "dummy" names in now there's a real one?

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,325 @@
...
+        ]>
+
+<sconsdoc xmlns="http://www.scons.org/dbxsd/v1.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://www.scons.org/dbxsd/v1.0 http://www.scons.org/dbxsd/v1.0/scons.xsd">
+
+    <tool name="ninja">
+        <summary>
+            <para>
+                Sets up &b-link-Ninja; builder which generates a ninja build file, and then optionally runs ninja.
```
mwichmann: when referring to the ninja command it might be worth annotating it... unfortunately `<application>` doesn't render with our current settings, so probably `<command>ninja</command>` or add it to `doc/scons.mod` and use `&ninja;`.

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,325 @@
...
+
+        </uses>
+    </tool>
+
+    <builder name="Ninja">
+        <summary>
+            <para>
+                &b-Ninja; is a special builder which
+                adds a target to create a ninja build file.
+                The builder does not require any source files to be specified,
```
mwichmann: nit: trailing comma - intend period?

SCons/Tool/ninja/ninja.xml
```diff
@@ -0,0 +1,325 @@
...
+            </para>
+        </summary>
+    </builder>
+
+
+    <cvar name="NINJA_GENERATED_SOURCE_SUFFIXES">
+        <summary>
+            <para>
+                The list of source file suffixes which are generated by SCons build steps.
+                All source files which match these suffixes will be added to the _generated_sources alias in the output
```
mwichmann, line 186: since this is prelim we can fixup these later - annotation of filenames and the alias name

SCons/Script/SConsOptions.py
bdbaddog: Yes. Easter eggs are fun?

SCons/Tool/ninja/ninja.xml
bdbaddog: addressed in next push.

PR #3652, opened 2020-05-14
fix travis coverage

dmoody256:
travis coverage is failing to generate report primarily from the move of SCons out of src dir. This PR also cleans up some other things which were messy and not necessary now that the tests finish much quicker.

Appveyor coverage seems to be working, but it could use a clean up. This PR is trying to focus on travis.


bdbaddog:
Looks good.
Merging!

PR #3659, opened 2020-05-18
let ZipBuilder actually use ZIPCOMSTR

RoccoMatano:
Currently the only way to set ZIPCOMSTR in such a way that it is actually used is to do so when creating an environment:
`env = Environment(ZIPCOMSTR = "Zipping $TARGET")`

Unlike other *COMSTR (like CXXCOMSTR) setting ZIPCOMSTR after creating an environment has no effect:
```
env = Environment(tools=["zip"])
env["ZIPCOMSTR"] = "Zipping $TARGET"
```

With this change it does.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,8 +10,11 @@ NOTE: Please include a reference to any Issues resolved by your changes in the b
 
 RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
+  From Rocco Matano:
+    - Make setting ZIPCOMSTR actually have an effect.
+
   From Daniel Moody:
-    - Add no_progress (-Q) option as a set-able option. However, setting it in the 
+    - Add no_progress (-Q) option as a set-able option. However, setting it in the
       SConstruct/SConscript will still cause "scons: Reading SConscript files ..." to be
       printed, since the option is not set when the build scripts first get read.
     - Added check for SONAME in environment to setup symlinks correctly (Github Issue #3246)
@@ -44,9 +47,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Add msys2 installed mingw default path to PATH for mingw tool.
       - C:\msys64\mingw64\bin
     - Purge obsolete internal build and tooling scripts
-    - Allow user specified location for vswhere.exe specified by VSWHERE. 
+    - Allow user specified location for vswhere.exe specified by VSWHERE.
       NOTE: This must be set at the time the 'msvc' 'msvs' and/or 'mslink' tool(s) are initialized to have any effect.
-    - Resolve Issue #3451 and Issue #3450 - Rewrite SCons setup.py and packaging. Move script logic to entry points so 
+    - Resolve Issue #3451 and Issue #3450 - Rewrite SCons setup.py and packaging. Move script logic to entry points so
       package can create scripts which use the correct version of Python.
     - Resolve Issue #3248 - Removing '-Wl,-Bsymbolic' from SHLIBVERSIONFLAGS
       NOTE: If your build depends on the above you must now add to your SHLIBVERSIONFLAGS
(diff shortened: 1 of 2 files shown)


bdbaddog:
You've ignored the pull request template.  I've re-added it above.
Please add a blurb to CHANGES.txt and add a test.

You can check the documentation as there's no need to update it.

RoccoMatano:
I shortly thought about extending the Zip tests, but had to realize that the effort for learning yet another unittest framework exceeds both my interest in this PR as well as my time budget.
Sorry for having bothered you.

bdbaddog:
@RoccoMatano - We should be able to take the work from here if you're unable. Please minimally add a blurb to CHANGES.txt so we can give you credit for finding and fixing this issue.

RoccoMatano:
I guess the test scons/test/ZIP/ZIPCOMSTR.py does not work reliably. If it would, this simple script
```
env = Environment(tools=["zip"], ZIPCOMSTR='Zipping $TARGET from $SOURCE')
env.Zip('test.zip', 'zip_scons.py')
```
should not print
```
>scons -f zip_scons.py
scons: Building targets ...
zip(["test.zip"], ["zip_scons.py"])
scons: done building targets.
```

But I don't understand what the test is doing. What I do understand is, that scons/SCons/Tool/zip.py
first creates one builder with the '$ZIPCOM' action
```
ZipBuilder = SCons.Builder.Builder(action = SCons.Action.Action('$ZIPCOM', '$ZIPCOMSTR'),
                                   source_factory = SCons.Node.FS.Entry,
                                   source_scanner = SCons.Defaults.DirScanner,
                                   suffix = '$ZIPSUFFIX',
                                   multi = 1)
```
and later sets `env['ZIPCOM'] = zipAction` . And `zipAction` is missing a 'COMSTR' (as the second parameter)
`zipAction = SCons.Action.Action(zip, varlist=['ZIPCOMPRESSION'])`
In my opinion it is that missing parameter, that causes the output I am seeing. That's why I added it in this PR.

Anyway, I will add a line to CHANGES.txt.

bdbaddog:
> ...
>. And `zipAction` is missing a 'COMSTR' (as the second parameter)
> `zipAction = SCons.Action.Action(zip, varlist=['ZIPCOMPRESSION'])`
> In my opinion it is that missing parameter, that causes the output I am seeing. That's why I added it in this PR.
> 
> Anyway, I will add a line to CHANGES.txt.

An output string is not required. 

From the manpage in the "Action objects" section:

> The second argument is optional and is used to define the output which is printed when the Action is actually performed. In the absence of this parameter, or if it's an empty string, a default output depending on the type of the action is used. For example, a command-line action will print the executed command. The argument must be either a Python function or a string.

RoccoMatano:
> An output string is not required.

Sure it's not. But it it is left out SCons will construct the the output string itself from the information it has. Exactly as can be seen in my example above:

>zip(["test.zip"], ["zip_scons.py"])

The whole point of letting a user set ZIPCOMSTR is to let her customize that output. And that's what is not working.

bdbaddog:
> > An output string is not required.
> 
> Sure it's not. But it it is left out SCons will construct the the output string itself from the information it has. Exactly as can be seen in my example above:
> 
> > zip(["test.zip"], ["zip_scons.py"])
> 
> The whole point of letting a user set ZIPCOMSTR is to let her customize that output. And that's what is not working.

Ahh. I see that now. 
O.k. one suggestion. 
If you find other issues and have a solution, but no time to complete a full pull request, Just file an issue with your patch?  Then we can file a PR and reference your issue and patch.
A little simpler for the maintainer.
Much thanks for bringing this issue to the projects attention!

mwichmann:
FWIW all the CI builds that "failed" triggered intermittent things that are known, and do not seem to have been caused by this PR.

bdbaddog:
Closing.
I'll be filing a follow up PR with a test and a little better explanation of the problem.

```  
Fix Zip tool to respect ZIPCOMSTR. Previously all zip builder calls would yield something
like zip(["test.zip"], ["zip_scons.py"]) and ignore ZIPCOMSTR if ZIPCOM and ZIPCOMSTR
weren't set after the Environment/Tool is initialized. (Explained in PR #3659)
```

Issue #3247, opened 2018-11-28
Introduce new variable SOVERSION to improve on SONAME


acmorrow:
* Link to SCons Users thread discussing your issue.
Mentioned, though not discussed here: https://pairlist4.pair.net/pipermail/scons-users/2018-November/007441.html

* Version of SCons
All offering SONAME support

Right now, if you use `SONAME`, you need to put the `LIBPREFIX` and `LIBSUFFIX` on yourself. That is fragile. It would be better to introduce a `SOVERSION` option that would then get properly interpolated into the filename used for `SOVERSION`.

bdbaddog:
Can you include a small example sconstruct?

acmorrow:
@bdbaddog - As this is an enhancement (there is currently no `SOVERSION` variable), I don't think this is something urgent to get done for SCons 4.0. I think #3246 is something we should see in 4.0 though.

acmorrow:
@bdbaddog - Could we queue this up for a 4.1? I don't think it is hard.

dmoody256:
I would assume that if the two are used together in the same environment, the SONAME takes precedence (assuming the end user may really want to make a weirdly named library symlink)?

acmorrow:
I think maybe it should just be an error to have both set. If you set `SONAME` explicitly then you get what you asked for, and also setting `SOVERSION` doesn't make sense.

One nice way to handle this might be to introduce a `SONAME_GENERATOR` configurable. The default implementation would work like if `SONAME` is set in the environment, then `SONAME_GENERATOR` just returns that. Otherwise, if `SOVERSION` is set, then it does the platform appropriate interpolation.

bdbaddog:
@acmorrow @dmoody256 - Isn't this resolved in 4.1.0? With the versioned shared library rewrite to pull in also the changes from @dmoody256 's PR for SOVERSION

dmoody256:
@bdbaddog yeah this is done we can close it

bdbaddog:
Closing

Issue #3246, opened 2018-11-28
InstallVersionedLib should respect SONAME variable


acmorrow:
As discussed on the mailing list, InstallVersionedLib doesn't respect SONAME when set, and probably should.

* Link to SCons Users thread discussing your issue.
https://pairlist4.pair.net/pipermail/scons-users/2018-November/007441.html

* Version of SCons
All that offer InstallVersionedLib

* What Platform are you on? (Linux/Windows and which version)
Mac/Linux. Windows doesn't really have these concepts.

acmorrow:
@bdbaddog - I think it is worth trying to fix this for 4.0. The `SONAME` is a documented SCons Variable, and it shouldn't become ignored just because you use `SHLIBVERSION` with `InstallVersionedLib`. I think we can probably live without #3247 for 4.0. While unfortunate, there is a user-facing workaround. However, this issue does not have one that I can see.

acmorrow:
@bdbaddog - I think this is worth fixing for 4.0 too. It makes using InstallVersionedLibrary impossible if you are using a SONAME scheme where ABI is versioned independently of release versioning.

bdbaddog:
Fixed by #3635

PR #3661, opened 2020-05-19
[WIP] Add SONAME_GENERATOR and SOVERSION to allow specifying custom logic to name versioned shared libraries

dmoody256:
added alternate ways to set library .so symlink versions from feature request: #3247 

Overview:

The SOVERSION is used by the default SONAME_GENERATOR to place the version number into the generated symlink for the library. For example, a target library name 'foo' and SOVERSION='4', build a symlink of ```libfoo.so.4``` on linux, and ```libfoo.4.dylib```. example code:

```
SharedLibrary(
  'foo',
  'foo.c',
  SHLIBVERSION='0.1.2',
  SOVERSION='4')
```

The SONAME_GENERATOR allows you to replace the default SONAME_GENERATION with a custom class. The call function must be pass the Environment and Node of the library. It should return a string which assembles what the library .so name which links to the real library, allowing the user complete customization of the .so name. The example below will build ```libMyCustomSoname.so.4```

```
class MySoNameGenerator(object):
    def __init__(self):
        pass
    def __call__(self, env, libnode, *args, **kw):
        return '%sSoname.so.%s' % (
          env.get('SONAME', 'lib'), 
          env.get('SOVERSION', '0'))

SharedLibrary(
  'test',
  'test.c',
  SHLIBVERSION='0.1.2',
  SONAME='MyCustom',
  SOVERSION='2',
  SONAME_GENERATOR = MySoNameGenerator())
```

## Contributor Checklist:


SCons/Tool/__init__.py
```diff
@@ -687,8 +687,16 @@ def __call__(self, env, libnode, **kw):
             print("_LibSonameGenerator: libnode=%r" % libnode.get_path())
 
         soname = _call_env_subst(env, '$SONAME', **kw2)
+        version = _call_env_subst(env, '$SOVERSION', **kw2)
```
bdbaddog, line 697: Perhaps check if the values are not equal?
Also dump the values in the error message.

SCons/Tool/__init__.py
```diff
@@ -707,11 +715,6 @@ def __call__(self, env, libnode, **kw):
 
         return soname
 
-
-ShLibSonameGenerator = _LibSonameGenerator('ShLib')
-LdModSonameGenerator = _LibSonameGenerator('LdMod')
```
bdbaddog, line 712: Why remove these?

SCons/Tool/applelink.py
```diff
@@ -71,12 +68,7 @@ def _applelib_versioned_lib_soname(env, libnode, version, prefix, suffix, name_f
         print("_applelib_versioned_lib_soname: name={!r}".format(name))
     major = version.split('.')[0]
     (libname,_suffix) = name.split('.')
-    # if a desired SONAME was supplied, use that, otherwise create
```
bdbaddog, line 74: Why remove this?

SCons/Tool/dmd.py
```diff
@@ -136,7 +136,7 @@ def generate(env):
     env['_SHDLIBSONAME'] = '${DShLibSonameGenerator(__env__,TARGET)}'
     # NOTE: this is a quick hack, the soname will only work if there is
     # c/c++ linker loaded which provides callback for the ShLibSonameGenerator
-    env['DShLibSonameGenerator'] = SCons.Tool.ShLibSonameGenerator
+    env['DShLibSonameGenerator'] = env.get("SONAME_GENERATOR", SCons.Tool._LibSonameGenerator)('ShLib')
```
bdbaddog: This means you can only alter this behavior by setting SONAME_GENERATOR when the tool is initialized?
Likely that's not the desired functionality.
You should be able to do this when calling ```env.SharedLibrary(target, sources, SHNAME_GENERATOR=special_one)```

bdbaddog:
This comment https://github.com/SCons/scons/pull/3661/files#r432688132 applies several places where you have the same logic.

SCons/Tool/applelink.py
dmoody256: This was a fix to #3246, but it was not the correct way to fix it, because ShLibSonameGenerator is meant to call the function it was called in.

bdbaddog:
Can you add a blurb to the description on how your changes are supposed to work?

SCons/Tool/__init__.py
dmoody256: These were no longer used so i removed them. However I reverted the changes to the D tools so these are no longer removed in d9c4217.

SCons/Tool/applelink.py
acmorrow: I'm not sure what you mean by "call the function it was called in". Can you elaborate?

acmorrow:
I'm having a difficult time seeing how all the pieces of this fit together. Could you provide an overview and description how this works?

dmoody256:
@bdbaddog @acmorrow I updated the description. There is a few more things I need to work out with this soname generator as well. Marking WIP.

SCons/Tool/applelink.py
dmoody256: Sorry, when I responded to this I got ripped away from the computer before I could write it all out. 

The ShLibSonameGenerator (really a specific instance of the base class _LibSonameGenerator for shlibs) is the base generator of sonames. If you look at the class, when there is no SONAME defined, the _LibSonameGenerator will attempt to call the generate_versioned_lib_info, which then gets and calls a specific callback from env['LINKCALLBACKS'], which in in applelink would be this function.

The real issue from #3246 was in ```_versioned_shlib_symlinks``` (updated in this PR) which should be setting soname_func as ShLibSonameGenerator, was just skipping down to the LINKCALLBACK for the specific type of library. That function should have let ShLibSonameGenerator decide to call the LINKCALLBACK if the SONAME was not defined, instead of the specific library function making the decision.

bdbaddog:
so the title of the PR is really not useful. Can you change it to something informative?

I'm thinking "Add SONAME_GENERATOR to allow specifying custom logic to name versioned shared libraries" ? assuming I now understand the purpose?

dmoody256:
Broke out the SOVERSION part of this, and will attack the custom SONAME_GENERATOR part in another PR. Closing this PR as its convoluting the two features.

PR #3663, opened 2020-05-20
Fix coverage reports

dmoody256:
recently #3652 fixed generating reports, but it deleted the coverage file environment variable which causes some of the coverage reports to get lost. This PR returns the coverage file.

This adds the coverage file back in and also cleans up appveyor aligning it with travis, removing linux jobs and build jobs.

Example of the coverage report from my branch:  
https://codecov.io/gh/dmoody256/scons/tree/aa5378e37ff2ab4e70be6d6b2af0cd79597da7dc

## Contributor Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,31 +67,7 @@ jobs:
       python: 3.7
       name: coverage
       before_script:
-        - python -m pip install -U coverage codecov
-        - python -m site
-        # setup sitecustomize so we can make all subprocess start coverage
-        - export PYSITEDIR=$(python -c "import sys; print(sys.path[-1])")
-        - export COVERAGE_PROCESS_START="$PWD/.coveragerc"
-        - export COVERAGE_FILE="$PWD/.coverage"
-        - mkdir -p "$PYSITEDIR"
-        - echo "$PYSITEDIR"
-        - touch "${PYSITEDIR}/sitecustomize.py"
-        - echo "import os" |  tee --append "${PYSITEDIR}/sitecustomize.py"
-        - echo "os.environ['COVERAGE_PROCESS_START'] = '$PWD/.coveragerc'" | tee --append "${PYSITEDIR}/sitecustomize.py"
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -72,23 +72,25 @@ jobs:
         # setup sitecustomize so we can make all subprocess start coverage
         - export PYSITEDIR=$(python -c "import sys; print(sys.path[-1])")
         - export COVERAGE_PROCESS_START="$PWD/.coveragerc"
+        - export COVERAGE_FILE="$PWD/.coverage"
         - mkdir -p "$PYSITEDIR"
```
bdbaddog: any chance I can convince you to just make this another shell script in .travis
And/or update the .coveragrc which is already in .travis?

dmoody256: yeah I can move this into shell scripts. The main reason for not just having it all static in a .coveragerc was so the absolute path from $PWD would make it into the .coveragerc file

dmoody256: maybe a shell script could generate the .coveragerc with some replace logic for the PWD? Would be cleaner then having everything in a script

bdbaddog: I think the powershell steps can be scripted as well.

Example from another repo where they do:

```
build_script:
  - ps: .\build.ps1 --bootstrap
  - ps: .\build.ps1 -Target ContinuousIntegration
```

https://github.com/chocolatey/ChocolateyGUI/blob/develop/.appveyor.yml
https://github.com/chocolatey/ChocolateyGUI/blob/develop/build.ps1

Seems like having many lines of shell logic in the .travis and .appveyor is hard to read/maintain/debug..

.coveragerc
```diff
@@ -0,0 +1,30 @@
+[run]
```
bdbaddog: Probably don't want to check this one in?

bdbaddog:
Nice work. Much cleaner!
1 note see annotations.

dmoody256:
Looks good now, the appveyor failure was test\Interactive\configure.py.

PR #3670, opened 2020-05-26
Added flags for SConscript nodes to be easily identified

dmoody256:
possible implementation for #3625 

Adds a global set recording all SConstruct/SConscript nodes, also added some functions to SCons.FS, get_SConstructs and get_SConscripts.

Added the .is_sconscript function to the base node class.

## Contributor Checklist:


bdbaddog:
Added [WIP] to indicate still a work in process. when you think it's done, remove the [WIP].

SCons/Script/SConscript.py
```diff
@@ -202,6 +202,7 @@ def _SConscript(fs, *files, **kw):
                 else:
                     f = fs.File(str(fn))
                 _file_ = None
+                f.is_sconscript = True
```
bdbaddog: You'd need to add this to __slots__

SCons/Node/FS.py
```diff
@@ -1199,8 +1200,9 @@ def __init__(self, path = None):
         DirNodeInfo.fs = self
         FileNodeInfo.fs = self
 
-    def set_SConstruct_dir(self, dir):
+    def set_SConstructs(self, dir, scripts=None):
         self.SConstruct_dir = dir
+        self.SConstructs = scripts
```
dmoody256: this change was somewhat specific for the ninja tool, which wants to know the SConstruct files specifically. Technically SConstructs_dir can be computed from Node.SConstructs[0], but I left the original attribute for backwards compatibility

bdbaddog:
I'd say just create a global list of nodes which are SConscripts rather than annotating all such nodes and/or adding a slot for the info.

bdbaddog:
And an API to retrieve the info.

dmoody256:
> I'd say just create a global list of nodes which are SConscripts rather than annotating all such nodes and/or adding a slot for the info.

The trade off being less memory usage vs speed of iterating a list each time you want to check a given node is an SConstruct?

SCons/Script/SConscript.py
dmoody256: I added it to the ```__slots__``` in SCons.Node.Node class which is the base node class. Was there more to update?

bdbaddog:
I'm o.k with Ninja output taking that hit I think.
Though maybe better using node.attribute.is_sconscript rather than node.is_sconscript.

dmoody256:
I'll try to generate some metrics of memory vs build time when I get a chance for
1. using pure slot
2. using node's attributes slot
3. using a global list

However I would guess using the nodes attributes slot is a good trade off

bdbaddog:
I'd rather not add memory usage to every Node object which is only necessary for a single builder.
If you use a set rather than a list checking for membership will be fast and I don't think order is important?

dmoody256:
okay I did a little bit of analysis, its a bit informal since I don't have much time but here are my results:

**Memory**
First I measured memory usage of having a slot vs not having a slot. I wrote this sconstruct for generating a fair amount of nodes:
```python
for i in range(100000):
    File('test_' + str(i))
```

I ran scons --debug=memory 20 times in a row on that SConstruct and took the average of memory values of the "memory after build" for both master and this PR. I calculated about a **0.5% increase** in memory for adding an extra slot vs not.

Not insignificant I would say, but not terrible.

**Speed**
I first measured the speed of accessing the slot for all nodes with this SConsctruct:

```python
import time

files = []

for i in range(100000):
    files.append(File('test_' + str(i)))

start = time.perf_counter()
for file_node in files:
    if file_node.is_sconscript:
        pass
stop = time.perf_counter()
print(str(stop - start))
```
which gave me this time:
**slot time: 0.01244 seconds**

Next I tested using nodes.attributes.is_sconscript:
```python
import time

files = []

for i in range(100000):
    files.append(File('test_' + str(i)))

start = time.perf_counter()
for file_node in files:
    if hasattr(file_node.attributes, 'is_sconscript'):
        pass
stop = time.perf_counter()
print(str(stop - start))
```
which gave me this time:
**node.attributes time: 0.04917 seconds**

Finally I tried using the set list of sconstructs nodes:
```python
import time
import random

files = []
sconscripts = []

for i in range(100000):
    files.append(File('test_' + str(i)))

for i in range(10):
    sconscripts.append(files[random.randrange(1, 100000, 1)])

sconscripts_set = set(sconscripts)

start = time.perf_counter()
for file_node in files:
    if file_node in sconscripts_set:
        pass
stop = time.perf_counter()
print(str(stop - start))
```
which gave me this:
**sconscripts set time: 0.01404 seconds**

**Conclusion**
Putting the sconscripts in a set and accessing them later seems the best option, its almost as quick as slots on each node, and its almost no extra memory usage.

Also have the sconscripts all in a set, has the bonus that at anytime a user may get all the sconscripts nodes very quickly instead of needing to walk the DAG and look at all the nodes.

bdbaddog:
Actually your comparison is slightly flawed as you'd have to do a tree walk to find all the nodes and only return those which are SConscripts. So in that case I'd expect the perf difference to be even higher.

Though I was thinking some kind of bit flag on nodes to indicate certain special nodes could be useful. But as with most things I like the simplest approach first.. then optimize lest we forget 

> "premature optimization is the root of all evil."  -  quote by Sir Tony Hoare (popularized by Donald Knuth)

I do appreciate your diligence in finding the best solution via code.. :)

dmoody256:
> Actually your comparison is slightly flawed as you'd have to do a tree walk to find all the nodes and only return those which are SConscripts

Well my sconstructs make it a little unclear what I am testing, this is really a test given an input node, how long does it take to determine that node is an sconscript node or not. I do the loop to get an accumulated total which can be compared across the three timing cases, similar to an average. It may have been more clear if I wrote the sconstructs like this:

```python
import time
import random

files = []
sconscripts = []

for i in range(100000):
    files.append(File('test_' + str(i)))

for i in range(10):
    sconscripts.append(files[random.randrange(1, 100000, 1)])

sconscripts_set = set(sconscripts)

# function under test
def some_function(node):
    if node in sconscripts_set:
        # do something special for sconscripts
        pass

start = time.perf_counter()
for file_node in files:
    some_function(file_node)
stop = time.perf_counter()
print('Average time: {:.6f} ms'.format(((stop - start)/100000.0)*1000.0))
```

None the less I am happy with the results.

bdbaddog:
Indeed your testing proves that the other two methods in the simplest case aren't significantly faster than the global set().

A full implementation would have to walk all the nodes in SCons's Node tree looking for ones which were marked. Don't know if we have a general tree walker API (even internally).. 
Anyway.. it's all an aside.

Good work!

dmoody256:
I added some new functions to FS class and Node class, but I am not sure at what level functions need to be added to the documentation? There are many functions a user could call that are not in the documentation so it wasn't obvious to me if these functions fall into those categories? I am fine adding it, if someone could give me a hint to whats expected that would be great.

SCons/Node/FS.py
```diff
@@ -1199,9 +1201,31 @@ def __init__(self, path = None):
         DirNodeInfo.fs = self
         FileNodeInfo.fs = self
 
-    def set_SConstruct_dir(self, dir):
+    def set_SConstructs(self, dir, scripts=None):
+
```
bdbaddog: Add a docstring

dmoody256: followed your advice from discord and just removed the unused functionality

PR #3703, opened 2020-06-17
Integrate the stubprocess.py wrapper from Parts

dirkbaechle:
With this PR we integrate the stubprocess.py wrapper from Parts to our project. It's made available as an experimental feature (for now) via the '--stubprocess-wrapper' command line option.

I considered to simply always activate the wrapper, but the following reasons speak against it:

1. It can slow down builds with low memory consumption and/or low number of files. See also the 'man page' entry for this feature in 'doc/man/scons.xml'.
2. Python itself is now adding support for vfork starting in v3.7.2, but only for 'linux' platforms at the moment. However, a few Python releases from now this wrapper might be superfluous.
3. I had to patch the original version a lot to make it work under Python3, but I couldn't test my changes under OSx.
4. So far, Python3 has changed the interface for the subprocess._execute_child() call several times, and the wrapper should handle all of those at the moment. But as soon as Python changes again, users' builds would break immediately and no work-around (except patching the stubprocess wrapper) would be available.

I ran the wrapper against several 'artificial' builds (created with a Perl script by E. Melski) with the following results (all in parallel with '-j 4'):

```
Small (approx 6000 files)
-------------------------

normal:    1m4.336s
wrapper:   1m38.203s

Large (approx 78000 files)
--------------------------

normal:   14m33.155s
wrapper:  14m6.347s

Very large (approx 400000 files)
--------------------------------

normal:  406m9.926s
wrapper: 111m6.048s
```

There are no additional test cases provided with this PR, running the whole testsuite without the '--stubprocess-wrapper' option didn't show any additional fails on my machine. I also verified that the wrapper works fine in "interactive" mode.

For this current implementation of the wrapper, a Python3 floor version of 3.4 is assumed.

After this PR gets integrated, I plan to do a little report on the scaling of the wrapper over the numbers of cores and the number of files to build. I will then work my results into the 'WhySConsIsNotSlow' Wiki page and update it accordingly.

## Contributor Checklist:

### DIFF
--- a/doc/man/scons.xml
+++ b/doc/man/scons.xml
@@ -1220,6 +1220,7 @@ command:</para>
 -n, --no-exec, --just-print, --dry-run, --recon
 -Q
 -s, --silent, --quiet
+--stubprocess-wrapper
 --taskmastertrace=FILE
 --tree=OPTIONS
 </literallayout>
@@ -1667,6 +1668,32 @@ build process.
 The default value is to use a stack size of 256 kilobytes, which should
 be appropriate for most uses.  You should not need to increase this value
 unless you encounter stack overflow errors.</para>
+  </listitem>
+  </varlistentry>
+
+  <varlistentry>
+  <term><option>--stubprocess-wrapper</option></term>
+  <listitem>
+<para>Tries to activate the <filename>stubprocess.py</filename> wrapper from the <emphasis>Parts</emphasis>
+project for speeding up builds. This wrapper uses pickling of the Actions' arguments
+and calls itself again as a <emphasis>main</emphasis> Python script in order to let us use the fast
+<emphasis>vfork</emphasis> from the libc library. Currently, only the platforms
+<function>linux</function> and <function>darwin</function> are supported.</para>
+
+<para>Please note that the build time can actually grow when enabling this
+option, depending on the size of your project! Unfortunately, the redirection
+through a script lets us suffer from the startup times of the Python
+interpreter while searching and loading its required modules. So, if you have only a
+handful of files, all <emphasis>vfork</emphasis> benefits might get eaten
+up by the additional starting time for Python in each executed Action. As a result,
+your build gets slower in total.</para>
+
+<para>It's not possible to come up with a single recommendation, like: "If your project
+has more than 20000 files, use this option". Apart from the number of files to build,
+the memory consumed by the overall process, your OS and your HDD, it's also the Python interpreter
+version and which modules and dist-packages you have installed in it, that are all affecting
+the break-even point. In the end you simply have to try it out, and continue to use this
+option if you happen to find that it speeds up your build.</para>
   </listitem>
   </varlistentry>
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
   From Dirk Baechle:
     - Added Docker images for building and testing SCons. (issue #3585)
+    - Integrated stubprocess.py wrapper from the Parts project in order
+      to speed up large builds. It's available as an experimental option
+      '--stubprocess-wrapper' for now.
 
   From James Benton:
     - Improve Visual Studio solution/project generation code to add support
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -20,6 +20,12 @@
     - Add CompilationDatabase() builder in compilation_db tool. Contributed by MongoDB.
       Setting COMPILATIONDB_USE_ABSPATH to True|False controls whether the files are absolute or relative
       paths.  Address Issue #3693 and #3694 found during development.
+    - Added a new command-line option "--stubprocess-wrapper", which will try to wrap the
+      subprocess.Popen() call when activated. Enabling this option can give a significant speedup
+      for the build of large projects that consume a lot of memory and create many files.
+      Small builds however may experience a slow-down, which is why we declared this option
+      an "experimental feature" for the moment. Please also refer to the corresponding entry
+      in the MAN page for more information.
 
 
   DEPRECATED FUNCTIONALITY
--- a/SCons/Platform/__init__.py
+++ b/SCons/Platform/__init__.py
@@ -54,6 +54,9 @@ import SCons.Errors
 import SCons.Subst
 import SCons.Tool
 
+# Gets set to True in Main._main() when the '--stubprocess-wrapper'
+# option was given on the command-line for speeding up large builds.
+stubprocess_wrapper = False
 
 def platform_default():
     r"""Return the platform string for our execution environment.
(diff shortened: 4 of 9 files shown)


Jongy:
Hey @dirkbaechle, I posted a question to scons-users regarding SCons process spawning performance, and was referred to your PR (see https://pairlist4.pair.net/pipermail/scons-users/2020-June/008287.html).

I also wrote a PoC to tackle this issue: it's not as feature complete as Parts, but as far as I can tell, most of these features aren't used by SCons anyway (in `exec_subprocess()` and `exec_popen3()`). My current PoC replaces `exec_subprocess` which is the simplest case - only args and environment - and makes it send send "spawn requests" to a per-`Worker` spawner process. This alleviates most of the `fork()` and `execve()` memory mappings handling, and remains fairly cheap regarding the added syscalls.

I'd like to test it with the same benchmark, could you please publish the generator script here?

doc/man/scons.xml
```diff
@@ -1670,6 +1671,32 @@ unless you encounter stack overflow errors.</para>
   </listitem>
   </varlistentry>
 
+  <varlistentry>
+  <term><option>--stubprocess-wrapper</option></term>
+  <listitem>
+<para>Tries to activate the <filename>stubprocess.py</filename> wrapper from the <emphasis>Parts</emphasis>
+project for speeding up builds. This wrapper uses pickling of the Actions' arguments
```
grossag, line 1678: I found this confusing because it implied that Parts is also needed. If this truly is now part of SCons, I personally think that it would be more easily understandable if you either (1) dropped mention of the Parts project or (2) mentioned that it does not require Parts.

bdbaddog: Maybe add "a contribution from the Parts project"

bdbaddog:
Is there no way to create a test(s) for this?
Maybe repurpose some general tests to run with stubprocess?

dirkbaechle:
@bdbaddog Hmm, regarding the test: what would be the list of your acceptance criteria, i.e. what exactly would you like to test? (Sorry, I'm doing too much work with requirements in my day job at the moment. ;) )

dirkbaechle:
@Jongy Please check out https://bitbucket.org/dirkbaechle/scons_testresults/src/default/scons230_vs_make/ where I uploaded the script `scons_test.pl` a few minutes ago. I will also make the script available in a new version of my "scons_testsuite" repo (currently migrating from bitbucket to github), but this will take a while.

Jongy:
Thanks @dirkbaechle. I used your script to generate an environment this way:
```
# perl scons_test.pl -f 4000
Number of levels = 2
Number of dirs / level = 3
Number of source files / dir = 4000
Number of lookups / source file = 2
Number of compiles grouped      = 20
Expecting:
    directories:   12
    source files:  16,000
    include files: 16,004
    makefiles:     4 (1 per directory)
    total files:   32,011
    look-ups: >=   1,600
    rules: >=      840
When the build runs, 16,000 object files & 800 executable(s)
will be created, for a total of 48,811 files.
```

And these are the results on my Linux box, in a `python:latest` Docker (Python 3.8.3), running `scons -j 8`:
```
master @ e4fdc10cad4060c740763148699131be2dae3b96
real    3m6.849s
user    4m43.750s
sys     4m19.206s

process-spawner @ 9b411285fb0535376edcfc86a3cbc945c8da9b3e
real    1m40.905s
user    5m13.892s
sys     1m56.040s

stubprocess-wrapper @ 7c598af9a5b1a0ef4be618b9b56a5fb9a5ee9679
real	18m25.090s
user	48m20.900s
sys	89m10.878s
```

I ran with `perf` to identify the reasons for the unexpectedly long execution time for `stubprocess`: the `close` syscall is high in the graph, and so is seccomp filtering on syscalls (Docker adds the this seccomp overhead, per syscall).

Why so many `close`s? `SC_OPEN_MAX` appears to be `1048576`, and `os.closerange` on Linux just loops over fds and issues a million syscalls per `subprocess.Popen`... (while the C-side of `subprocess.Popen` has smarter logic for closing fds - it operates on open fds only, according to `/proc/pid/fd`.)

I then capped `MAXFD` at `1024` (sounds a fair default):

```
--- a/SCons/Platform/stubprocess.py
+++ b/SCons/Platform/stubprocess.py
@@ -87,7 +87,7 @@ if sys.platform in ('linux', 'darwin'):
 
     if __name__ == '__main__':
         try:
-            MAXFD = os.sysconf("SC_OPEN_MAX")
+            MAXFD = 1024
         except Exception:
             MAXFD = 256
```

and tried again (this time with `perl scons_test.pl -f 2000`):

```
master
real	1m22.913s
user	2m23.997s
sys	1m51.569s

process-spawner
real	0m48.841s
user	2m37.460s
sys	0m59.696s

stubprocess-wrapper
real	1m25.625s
user	7m11.600s
sys	1m54.429s
```

Given those results, I think the approach I presented in `process-spawner` should be taken into consideration, either as a substitute to this PR, or as a complementary, third option (so we have standard `stubprocess.Popen`, using Parts and using the process spawner)

dirkbaechle:
@Jongy If you have a solution that is even faster than the stubprocess.py, I'd say go for it. I'm personally not stuck to stubprocess.py...I simply took what the Parts project had to offer. One of my main goals wasn't to optimize the wrapper itself, but simply to make it work under SCons while modifying it as little as possible. Like this, it would be easier to keep both of those versions in synch, contributing changes and bugfixes back and forth.

What I would be interested in, are times of your wrapper with a lower number of files (6000 or even down to 4000 in total). Is it fast enough, such that we get faster build times even for small projects, or at least the same times roughly? Because then we could activate your wrapper as default. Just my 2cents.

Jongy:
Cool. Later today I'll post those additional results you've requested, and open a new pull request with my branch. I think that also for projects with very small number of files (e.g even < 100) the process spawning approach will be cheaper.

Jongy:
Perhaps it can be enabled by default... I made it configurable since I think that new, experimental features like this one shouldn't be enabled by default when first added, and definitely should be disable-able.

Following are the results for `perl /code/scons_test.pl -f 1000`. We can see that without `--process-spawner` the changeset didn't affect the times, but with `--process-spawner` we see some improvements.

The `real` is less relevant here, This test build doesn't run heavy compilations so CPU remains partly idle, and with `--process-wrapper` SCons becomes more concurrent and uses multiple cores more efficiently. What's more interesting is the reduction in total `user + sys` time, which means SCons requires less CPU time in total.

```
master @ ea848745d7025d8f6885dfbeaaa92731ddd711bd (I based my branch on this commit)
real	0m35.622s
user	1m13.946s
sys	0m40.853s

proces-spawner @ 758ba3d1e0a335c72e5489ec697a1a9af548fd90 (**without** --process-spawner)
real	0m35.371s
user	1m13.873s
sys	0m40.570s

proces-spawner @ 758ba3d1e0a335c72e5489ec697a1a9af548fd90 (**with** --process-spawner)
real	0m23.134s
user	1m18.697s
sys	0m28.367s
```

I'll submit a PR soon.

bdbaddog:
Based on notes from pr #3719 where py 3.10 will yield a big improvement without any scons changes, I'm going to close this PR. 

If there's still a reason to consider this work, please comment and reopen.

PR #3741, opened 2020-07-11
Add configurable dir to tempfilemunge vi TEMPFILEDIR

bdbaddog:
Allow configuring which directory TEMPFILEMUNGE creates it's temporary files in.
Added ```env['TEMPFILEDIR']```

## Contributor Checklist:


bdbaddog:
@dmoody256  - this should remove the bit in the generate about controlling tempfile generated file locations?

dmoody256:
> @dmoody256 - this should remove the bit in the generate about controlling tempfile generated file locations?

Yeah that should fix the bit about where controlling where tempfiles are created. 

Theres also a bit about controlling the deletion of the tempfile after the command is executed. Ninja will take the command list as to feed it in the generated file. Ninja handles tempfiles itself with response files, so it ends of overwriting TempFileMunge to just return the command.

Anyways this looks like it would be used for that last bit about setting the mkstemp location.

PR #3745, opened 2020-07-16
Fix fortran tools to set SHFORTRAN variables to $FORTRAN

datalogics-robb:
Fix fortran tools to set SHvariable to $variable (i.e. SHF77 to $F77 )
These changes address the problem noticed in issue #3740 with test CPPFLAGS.py.  This test assumed that changing the FORTRAN variable in an Environment would be inherited by SHFORTRAN.  That *should* be a valid assumption, but was not true for many of the fortran tools.  This PR corrects that, by setting SHFORTRAN to $FORTRAN much the way it is done in the C language tool.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,12 @@ NOTE: The 4.0.0 Release of SCons will drops Python 2.7 Support
 RELEASE 4.1.0.devyyyymmdd - Mon, 04 Jul 2020 16:06:40 -0700
 
   From Rob Boehne:
+    - Fix fortran tools to set SHFORTRAN variables to $FORTRAN, similarly SHF77, SHF90, SHF95,
+      SHF03 and SHF08 will default to the variables $F77, $F90, $F95, $F03 and $F08 respectively.
+      If you were depending on changing the value of FORTRAN (or $F[0-9][0-9]) having no effect
+      on the value of SHFORTRAN, this change will break that.   The values of FORTRAN, F77, F90,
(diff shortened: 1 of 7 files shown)


datalogics-robb:
Reviewer - is this behavior (default setting of SHFORTRAN) documented anywhere?

CHANGES.txt
```diff
@@ -9,6 +10,8 @@ NOTE: The 4.0.0 Release of SCons will drops Python 2.7 Support
 RELEASE 4.1.0.devyyyymmdd - Mon, 04 Jul 2020 16:06:40 -0700
 
   From Rob Boehne:
+    - Fix fortran tools to set SHFORTRAN variables to $FORTRAN (also for similar variables
```
bdbaddog: List all the tools you updated please. Add something like "If you were depending on setting FORTRAN not changing SHFORTRAN (or equivalent for fortran version you are using), this change will break that."

bdbaddog:
add the if 'SHF77' in env: to your logic as well.
The user should ALWAYS be able to override the default by specifying at the same time, or before the tool is initialized.

datalogics-robb:
Appveyor seems to have timed out one of the runs:

     Build execution time has reached the maximum allowed time for your plan (60 minutes).

bdbaddog:
restarted timed out appveyor

bdbaddog:
only failure is a spurious one on appveyor. 
Merging.

PR #3753, opened 2020-07-24
added SOVERSION for symlinked library generation from just a version number

dmoody256:
This PR covers just the SOVERSION updates from https://github.com/SCons/scons/pull/3661

The SOVERSION is a new environment variable which works very similar to the SONAME environment variable, except instead of providing the full library name as you do with SONAME, you just provide an number which will generally be the major version number in the symlinked library. 

The SHLIBVERSION environment variable is also very similar to this, except that the SHLIBVERSION is generally the specific version number of the library, and used on the actual output library, instead of the symlinked library.

Please review https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html for more information on the standard conventions for library versioning in linux.

Also note that macOS uses a similar convention for versioning libraries, so this PR also adds the variable to apple linker.

## Contributor Checklist:

### DIFF
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -448,7 +448,7 @@ class _ImpLibInfoSupport:
 
 class _LibInfoGeneratorBase:
     """Generator base class for library-related info such as suffixes for
-    versioned libraries, symlink maps, sonames etc. It handles commonities
+    versioned libraries, symlink maps, sonames etc. It handles commonalties
     of SharedLibrary and LoadableModule
     """
     _support_classes = {'ShLib': _ShLibInfoSupport,
@@ -673,10 +673,14 @@ class _LibSonameGenerator(_LibInfoGeneratorBase):
     """Library soname generator. Returns library soname (e.g. libfoo.so.0) for
     a given node (e.g. /foo/bar/libfoo.so.0.1.2)"""
 
-    def __init__(self, libtype):
+    def __init__(self, libtype=None):
         super(_LibSonameGenerator, self).__init__(libtype, 'Soname')
 
-    def __call__(self, env, libnode, **kw):
+    def __call__(self, env, libnode, *args, **kw):
+        # we can differentiate the type of this generator at the call time
+        # allowing us to decide this information during subst expansion
+        self.libtype = kw.get('libtype', self.libtype)
+
         """Returns a SONAME based on a shared library's node path"""
         Verbose = False
 
@@ -690,8 +694,15 @@ class _LibSonameGenerator(_LibInfoGeneratorBase):
             print("_LibSonameGenerator: libnode=%r" % libnode.get_path())
 
         soname = _call_env_subst(env, '$SONAME', **kw2)
+        version = _call_env_subst(env, '$SOVERSION', **kw2)
+        if soname and version:
+            raise SCons.Errors.UserError(
+                'Ambiguous library .so naming, both SONAME: %s and SOVERSION: %s are defined. '
+                'Only one can be defined for a target library.' % (soname, version))
+
         if not soname:
-            version = self.get_lib_version(env, **kw2)
+            if not version:
+                version = self.get_lib_version(env, **kw2)
             if Verbose:
                 print("_LibSonameGenerator: version=%r" % version)
             if version:
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -448,7 +448,7 @@ class _ImpLibInfoSupport:
 
 class _LibInfoGeneratorBase:
     """Generator base class for library-related info such as suffixes for
-    versioned libraries, symlink maps, sonames etc. It handles commonalties
+    versioned libraries, symlink maps, sonames etc. It handles commonalities
     of SharedLibrary and LoadableModule
     """
     _support_classes = {'ShLib': _ShLibInfoSupport,
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -673,14 +673,10 @@ class _LibSonameGenerator(_LibInfoGeneratorBase):
     """Library soname generator. Returns library soname (e.g. libfoo.so.0) for
     a given node (e.g. /foo/bar/libfoo.so.0.1.2)"""
 
-    def __init__(self, libtype=None):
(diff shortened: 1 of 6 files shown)


SCons/Tool/applelink.py
```diff
@@ -71,12 +69,7 @@ def _applelib_versioned_lib_soname(env, libnode, version, prefix, suffix, name_f
         print("_applelib_versioned_lib_soname: name={!r}".format(name))
     major = version.split('.')[0]
     (libname,_suffix) = name.split('.')
-    # if a desired SONAME was supplied, use that, otherwise create 
-    # a default from the major version
-    if env.get('SONAME'):
-        soname = ShLibSonameGenerator(env, libnode)
-    else:
-        soname = '.'.join([libname, major, _suffix])
+    soname = '.'.join([libname, major, _suffix])
```
bdbaddog, line 72: Why remove using supplied SONAME?

SCons/Tool/link.xml
```diff
@@ -378,11 +378,27 @@ See also &cv-link-LINKFLAGS;  for linking static objects.
...
+</cvar>
+
+<cvar name="SOVERSION">
+<summary>
+<para>
+Simlar to &cv-link-SONAME; but only modifies the version number, and not the name, suffix or prefix of a versioned shared library/loadable module.  This can not be used in the same environment as &cv-link-SONAME; because it would be ambiguous as to what the SONAME should be.
+<example_commands>
+env.SharedLibrary('test', 'test.c', SHLIBVERSION='0.1.2', SOVERSION='2')
+</example_commands>
+The variable is used by linker tools which use the &t-link-link; tool such as &t-link-gnulink; or &t-link-applelink;.
```
bdbaddog, line 404: Expand upon the SONAME generated when SOVERSION is specified. (what would it be in the example above)

SCons/Tool/link.py
```diff
@@ -243,17 +235,14 @@ def _versioned_lib_symlinks(env, libnode, version, prefix, suffix, name_func, so
 
 def _versioned_shlib_symlinks(env, libnode, version, prefix, suffix):
     name_func = env['LINKCALLBACKS']['VersionedShLibName']
-    soname_func = env['LINKCALLBACKS']['VersionedShLibSoname']
+    soname_func = env['ShLibSonameGenerator']
```
bdbaddog, line 238: Why change this?

SCons/Tool/applelink.py
dmoody256: This was not the correct order of the calling functions. ShLibSonameGenerator calls into LINKCALLBACK, which calls this function if SONAME is not supplied. Since ShLibSonameGenerator calls to the more specific naming functions, it should handle SONAME and SOVERSION in ShLibSonameGenerator, and then pass the naming info down.

SCons/Tool/link.py
dmoody256: ShLibSonameGenerator ends up calling that linkcallback if SONAME is not supplied. Without this change, the ShLibSonameGenerator is skipped out.

SCons/Tool/link.xml
dmoody256: done in 13527e8

SCons/Tool/__init__.py
```diff
@@ -673,10 +673,14 @@ class _LibSonameGenerator(_LibInfoGeneratorBase):
     """Library soname generator. Returns library soname (e.g. libfoo.so.0) for
     a given node (e.g. /foo/bar/libfoo.so.0.1.2)"""
 
-    def __init__(self, libtype):
+    def __init__(self, libtype=None):
         super(_LibSonameGenerator, self).__init__(libtype, 'Soname')
 
-    def __call__(self, env, libnode, **kw):
+    def __call__(self, env, libnode, *args, **kw):
```
dmoody256, line 679: this got pulled in from the SONAME_GENERATOR by accident, need to remove this change

dmoody256: removed in 027be11

dmoody256:
Ok tests are failing now that I removed that code I thought was left from the SONAME_GENERATOR. Maybe I took too much away, once I get a minute I'll go check it.

bdbaddog:
So if we allow setting SOVERSION (in place of setting SONAME) it should have the following impact:

1. the soname set in the shared library should be libname.SOVERSION
2. The chain of symlinks should be as below, where libname.so.SHLIBVERSION is the actual file.
```
   libname.so -> libname.so.SOVERSION
   libname.so.SOVERSION -> libname.so.SHLIBVERSION
``` 

  
3. For apple, the flag link flags:
`  gcc -o libfoo.SHLIBVERSION.dylib -dynamiclib -Wl,-current_version,SHLIBVERSION -Wl,-compatibility_version,SOVERSION.o foo.os`
4. For gnu the link flags should be
`gcc -o libfoo.so.SHLIBVERSION -shared -Wl,-soname=libfoo.so.SOVERSION foo.os`

In the case we have SONAME and not SOVERSION, should we extract the SOVERSION from SONAME for use in the ```-compatibility_version```?

Note we also have the following env vars
* APPLELINK_NO_CURRENT_VERSION - don't specify a -current_version
* APPLELINK_CURRENT_VERSION - use this for the -current_version and don't extract from SHLIBVERSION
* APPLELINK_NO_COMPATIBILITY_VERSION - don't specify a -compatibility_version
* APPLELINK_COMPATIBILITY_VERSION - use this for -compatibility_version and don't extract from SHLIBVERSION

The apple bits aren't really (yet) addressed in this PR.
(Note there's also some validation of the version numbers for apple. 0-65536[.0-255[.0-255]] are the only valid formats for those numbers and there's pre-existing code to check that)

Of note, some PR's for meson and links to cmakes implementation details

- https://github.com/mesonbuild/meson/issues/1451
- https://github.com/mesonbuild/meson/blob/master/test%20cases/osx/2%20library%20versions/CMakeLists.txt
- https://github.com/mesonbuild/meson/pull/4095
- https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

I'm guessing it's better for our user base if we match cmake/meson's implementations (rather than libtool, especially for apple).

bdbaddog:
Closing in favor of PR #3848 which is a rewrite of existing logic and implements this logic as well.

PR #3763, opened 2020-08-03
Fix SCons.Node.FS.File.get_csig() usage of md5_chunksize (Issue #3726)

### LINKED ISSUE #3726: Code in SCons.Node.FS.File.get_csig() checking md5_chunksize is not correct
A coworker noticed today that the code:

`if self.get_size() < File.md5_chunksize:`

is not correct because md5_chunksize is in kilobytes, not bytes. I don't have a good test case, but I'm filing this against myself to track fixing it after finding a good test case.

grossag:
md5_chunksize is documented to be in kilobytes, but this function was using it as though it is in bytes. This change fixes that, adding a test to confirm the correct behavior for a few different cases.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
         - Whatever John Doe did.
 
+  From Adam Gross:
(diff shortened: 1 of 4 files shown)


bdbaddog:
Wouldn't we be better off to store the byte value in md5_chunksize, than multiply it everywhere?
Then you only have to multiply in SCons/Script/Main.py line 1115

PR #3771, opened 2020-08-10
fix issue where java parses class incorrectly from lambdas after new

dmoody256:
Found a small issue related to parsing java files. I assume this was never run into in earlier versions of java because lambda was added in java 1.8.

With out the small change to see lambda tokens, the java parser would skip over that notation and think a anonymous class was being created if a new was used just prior. This change make the java parser aware of the lambda notation.

See the supplied test case for more info.

## Contributor Checklist:


SCons/Tool/JavaCommon.py
```diff
@@ -88,7 +88,7 @@
     #     the multi-line comment begin and end tokens /* and */;
     #     array declarations "[]".
     _reToken = re.compile(r'(\n|\\\\|//|\\[\'"]|[\'"{\};.()]|' +
-                          r'\d*\.\d*|[A-Za-z_][\w$.]*|<[A-Za-z_]\w+>|' +
+                          r'\d*\.\d*|[A-Za-z_][\w$.]*|->|<[A-Za-z_]\w+>|' +
```
bdbaddog: Please update the comment block above with the change

SCons/Tool/JavaCommonTests.py
```diff
@@ -484,6 +484,41 @@ def test_nested_anonymous_inner_classes(self):
...
+{
+
+    public void testFunc (int arg1, String arg2, Runnable lambda){
+    }
+    public LamdaExample()
+    {
+        testFunc(
+            5,
+            new String("test"),
+            () -> {}
```
bdbaddog, line 505: Stick a comment above the lambda to indicate that's what it is? (Not everyone's familiar with Java lambda syntax)

bdbaddog:
A couple minor nits, but otherwise looks good.
Once you fix those I'll be able to merge.

SCons/Tool/JavaCommonTests.py
dmoody256: updated in 1b2e8e9

SCons/Tool/JavaCommon.py
dmoody256: updated in 1b2e8e9

bdbaddog:
Great!
Thanks.
Merging.

PR #3776, opened 2020-08-11
Add check for pre-specified YACC to yacc tool's exist()

bdbaddog:
Fix exist()

## Contributor Checklist:

### DIFF
--- a/SCons/Tool/yacc.py
+++ b/SCons/Tool/yacc.py
@@ -162,6 +162,9 @@ def generate(env):
 
 
 def exists(env):
+    if 'YACC' in env:
+        return env.Detect(env['YACC'])
+
     if sys.platform == 'win32':
         return get_yacc_path(env)
     else:


acmorrow:
LGTM

SCons/Tool/yacc.py
```diff
@@ -162,6 +162,9 @@ def generate(env):
 
 
 def exists(env):
+    if 'YACC' in env:
```
acmorrow, line 165: Is it a good idea to include a check that `YACC` is something truthy as well?

acmorrow: Or will Detect just deal with that for you?

bdbaddog: If YACC is garbage Detect will be False (or throw an exception if it's not understandable garbage I suppose)

acmorrow: Sounds good. LGTM for sure then. Thanks for the quick fix.

bdbaddog:
Merging.

PR #3825, opened 2020-11-16
Teach the Python scanner to find generated files and directories

### LINKED ISSUE #3824: Python scanner does not support scanning dynamically generated nodes
**Describe the bug**

The Python scanner from the 'Python' tool is not aware of dynamically generated files because it does not use SCons.Node.FS.find_file. Instead, it looks on disk for the dependency at scanning time.

**Required information**
* Link to SCons Users thread discussing your issue: on today's discord - https://discord.com/channels/571796279483564041/612459756246007829/776833745495261214
* Version of SCons: 4.0.1
* Version of Python: 3.8.6
* How you installed SCons: forked and cloned
* What Platform are you on? (Linux/Windows and which version): Windows 10 x64

I'm still working on a reproducer. Current draft is here: https://github.com/grossag/scons/blob/topic/grossag/pythondynamicscan/test/Scanner/Python.py with dir fixture here: https://github.com/grossag/scons/tree/topic/grossag/pythondynamicscan/test/Scanner/Python but it runs into some other unrelated problems.

General idea is:
1. SConstruct copies a package in.
2. SConstruct runs a script that imports the package at its destination name.
3. SCons should understand that the script depends on the copy action.

scons --debug=stacktrace .

grossag:
This change teaches the Python scanner about generated files and directories. Previously the Python scanner only could find files or directories if they existed on disk at scan time. This change switches the code to use `SCons.Node.FS.find_file` to find imports, regardless of whether they exist on disk.

Fixes #3824 

## Contributor Checklist:

### DIFF
--- a/SCons/Scanner/Python.py
+++ b/SCons/Scanner/Python.py
@@ -116,33 +116,40 @@ def scan(node, env, path=()):
 
         module_components = search_string.split('.')
         module_joined = '/'.join(module_components)
-        print('%s - %s' % (module_joined, [str(s) for s in search_paths]))
-        node = SCons.Node.FS.find_file(module_joined, search_paths, verbose=True)
-        if node:
-            # The fact that we were able to find the node without appending .py
-            # means that this is a directory import.
-            nodes.append(env.Dir(node).File('__init__.py'))
 
-            # Take a dependency on all __init__.py files from all imported
-            # packages unless it's a relative import. If it's a relative
-            # import, we don't need to take the dependency because Python
-            # requires that all referenced packages have already been imported,
-            # which means that the dependency has already been established.
-            # XXX TODO: This part is broken and needs to be fixed.
-            if not is_relative and len(module_components) > 1:
-                import_dirs = module_components
-                for i in range(len(import_dirs)):
-                    init_path = '/'.join(module_components[:i+1] + ['__init__.py'])
-                    # TODO: Passing search_paths is not correct.
-                    init_node = SCons.Node.FS.find_file(init_path, search_paths, verbose=True)
-                    if init_node:
-                        nodes.append(init_node)
-        else:
-            node = SCons.Node.FS.find_file(module_joined + '.py', search_paths, verbose=True)
+        # For an import of "p", it could either result in a directory named
+        # p or a file named p.py. We can't do two consecutive searches for p
+        # then p.py because the first search could return a result that is
+        # lower in the search_paths precedence order. As a result, it is safest
+        # to iterate over search_paths and check whether p or p.py exists in
+        # each path. This allows us to cleanly respect the precedence order.
+        for path in search_paths:
+            paths = [path]
+            node = SCons.Node.FS.find_file(module_joined, paths, verbose=True)
             if node:
-                nodes.append(node)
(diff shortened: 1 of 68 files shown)


SCons/Scanner/Python.py
```diff
@@ -82,6 +82,40 @@ def find_include_names(node):
...
+    If the import is not found, returns a tuple containing (None, False, None).
+    Callers should check for failure by checking whether the first entry in the
+    tuple is not None.
+    """
+    for search_path in search_paths:
+        paths = [search_path]
+        # Note: if the same import is present as a package and a module, Python
+        # prefers the package. As a result, we always look for x/__init__.py
+        # before looking for x.py.
+        node = SCons.Node.FS.find_file(import_path + '/__init__.py', paths)
```
djh82, line 108: Looking through this PR since it's functionality we'd like!  Can Windows handle `/`?  The usual python, OS independent, way of handling paths is `os.path.join`?

grossag: Good question. I did validate that the scanner and tests work on Windows, so yes I believe it's okay. As far as I can tell, `SCons.Node.FS.find_file` supports `/` because that could come into play when other scanners are used. My understanding is that the C scanner reading a line like `#include <foo/hello.h>` would trigger a similar type of filename being passed to `find_file`.

That said, either way is fine. I just tried replacing all usage of `/` with `os.sep` and all tests succeeded. So I'm open to whatever people prefer.

bdbaddog: os.path.join() is probably "correct", but yes, both windows, python, and SCons can handle "/"

grossag: I'd want to avoid using os.path.join() because it has extra logic for absolute paths unrelated to the goal here. For example, the docs say:
```
If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.
```
I am fine replacing `/` with `os.sep` though. Just let me know if you all want me to do that.

SCons/Scanner/Python.py
```diff
@@ -92,10 +126,14 @@ def scan(node, env, path=()):
         # if the same header is included many times.
         node.includes = list(map(SCons.Util.silent_intern, includes))
 
-    # XXX TODO: Sort?
     nodes = []
     if callable(path):
         path = path()
+
+    # If there are no paths, there is no point in parsing includes in the loop.
+    if not path:
```
bdbaddog: Is this correct? what if the modules are located in the same directory as the file?

grossag: Yeah good point. I'll need to move it below and change it to:

if not search_paths:

bdbaddog: Any updates?

grossag: Oops, sorry! Just pushed an update with a fix.

Separately, do you want me to merge master into this branch to bring it up-to-date? Wasn't sure if it would make the diff more complicated or if it would be transparent.

grossag: Nevermind, I figured out how to merge master in this branch, so I have done that.

bdbaddog:
Sorry for the delay.
Merging.

PR #3854, opened 2021-01-04
[WIP] Bugfix/managan/latex scanner

### LINKED ISSUE #2972: TEX builder/logic: A missing file causes failure with poor error message
This issue was originally created at: 2014-08-26 08:49:48.
This issue was reported by: `managan`.
managan said at 2014-08-26 08:49:48
>If a Tex file inputs a missing file the builder errors out and the message is totally obscure. There is a logic flaw that results in a an uninitialized reference that needs to be fixed so we get the LaTeX error message instead of a
>failure in the builder.

managan said at 2014-08-26 09:54:26
>I put a patch for this issue in https://bitbucket.org/managan/scons_newglossary-fix 
> 
>The test is not working yet. I need to figure out how to run a test and distinguish a bad stderr result from a "good" stderr result.

bdbaddog said at 2015-09-30 19:23:55
>Fix typo in summary

managan:
This pull request aims to fix a case where the LatexScanner fails. 

The example test is when the source includes `\input{filename}` and the file filename is not present.  This file does not show up in `--tree=all` output and then we get an error from `pdflatex` when it can not find the file.

The scanner needs to be updated to report the file is in the source tree even when it is not present.

The source change avoids an exception being thrown when the builder tries to determine if the deck is a tex file or a latex file.

## Contributor Checklist:

### DIFF
--- a/SCons/Tool/tex.py
+++ b/SCons/Tool/tex.py
@@ -559,9 +559,9 @@ def is_LaTeX(flist,env,abspath):
             if srcNode is not None:
                 file_test = is_LaTeX(fileList, env, abspath)
 
-            # return on first file that finds latex is needed.
-            if file_test:
-                return file_test
+                # return on first file that finds latex is needed.
+                if file_test:
+                    return file_test
 
         if Verbose:
             print(" done scanning ",str(f))
--- a/test/TEX/missingInput.py
+++ b/test/TEX/missingInput.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+#
+# __COPYRIGHT__
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+
+__revision__ = "__FILE__ __REVISION__ __DATE__ __DEVELOPER__"
+
+r"""
+Test creation of a LaTeX document that uses \input{filename}.
+Check that SCons does not throw an exception when the 
+file is not found.
+
+
+Test courtesy Rob Managan.
+"""
+
+import TestSCons
+
+test = TestSCons.TestSCons()
+
+latex = test.where_is('latex')
+if not latex:
+    test.skip_test("Could not find 'latex'; skipping test.\n")
+
+test.write(['SConstruct'], """\
+import os
+
+env = Environment(tools = ['pdflatex', 'pdftex'])
+
+test = env.PDF(source='test.tex')
+""")
+
+test.write(['test.tex'],
+r"""
+\input{preamble.tex}
+
+\begin{document}
+\title{The uncountability of SCons features}
+\author{Adam Ehlers Nyholm Thomsen}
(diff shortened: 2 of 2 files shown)


managan:
This is tied to Fixes #2972

managan:
#2972

bdbaddog:
@managan  - Is this ready to go? Or still have some items outstanding?

bdbaddog:
@managan - just a ping to see if you're still working on this?

managan:
Hi,
Still trying to free up some time for this. We are now in GA but still doing things on the new house!

bdbaddog:
> Hi, Still trying to free up some time for this. We are now in GA but still doing things on the new house!

No worries. Just wanted to touch base. 

Enjoy the house process. ;)

bdbaddog:
@managan  - just a ping to see if you have time to complete this PR.

managan:
I am retiring at the end of November. I am putting this on my list of things to look at in December!

bdbaddog:
> I am retiring at the end of November. I am putting this on my list of things to look at in December!

Congratulations!
And thank you! :)

managan:
Turning 68 in a few weeks so it is time!!


*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-
Rob Managan
WCI/DP Division  LLNL
P.O. Box 808, L-170
Livermore, CA  94551-0808
Email: ***@***.***
Phone: 925-423-0903         FAX:   925-422-3389
Mobile: 925-341-9939

From: William Deegan ***@***.***>
Reply-To: SCons/scons ***@***.***>
Date: Monday, September 11, 2023 at 5:33 PM
To: SCons/scons ***@***.***>
Cc: Rob Managan ***@***.***>, Mention ***@***.***>
Subject: Re: [SCons/scons] [WIP] Bugfix/managan/latex scanner (#3854)


I am retiring at the end of November. I am putting this on my list of things to look at in December!

Congratulations!
And thank you! :)

mwichmann:
Came across this again closing down old browser tabs ;-)  This in concept is similar to some Fortran stuff I've been wrestling with. A lot of scanners work on the assumption that "if a scanned dependency is not found, don't add it to the dep list, and just move on, as if these were no problem" - you see the same pattern over and over, probably copied:
```py
            node = node.rfile()
            if not node.exists():
                return []
```

As far as I've been able to deduce, the root of this assumption is that you don't want to generate dependencies for system headers - if `/usr/include/stdio.h` is missing, *we* don't have a way to regenerate it, so there's no point in adding it as a dep, the build will fail anyway when the header isn't found by the compiler, and the user has to install something through their package manager to resolve it.  But we don't have any special way of determining that something is a system-provided file as opposed to a project-provided one. Thus - we just ignore the case of not finding a file that looks like it's required. This sort of works for C projects, actually.

The problem is, if the file could be generated, then we absolutely want a dependency on it, so SCons can force it to be built - normally before the thing that includes it, or we get build-ordering problems that manifest as missing files - there are a number of these in the issue tracker (Fortran, Java, SWIG, D off the top of my head).

Don't have a general answer for this class of problems, because we don't have enough information to make an accurate decision. In the Fortran case, it manifests with module files, so I'm trying to treat a module dependency separately and not apply the template above.  That's not necessarily pretty... Can we do something similar with latex? Is this in fact a case of the same problem?

kprussing:
I'm not sure what the problem this PR is trying to solve. I threw together a quick [gist](https://gist.github.com/kprussing/73bcd06446f00c7df2c29bf728d91af7) to see what happens (at least as far back as 4.7.0 which I happen to have a scons-local on the computer). Running with a missing input file that SCons knows how to generate correctly populates the build tree. But for a file that SCons does not know how to build, it is not present. The compilation properly stops as expected with
```bash
cd . && pdflatex -interaction=nonstopmode -recorder main.tex
This is pdfTeX, Version 3.141592653-2.6-1.40.26 (TeX Live 2024) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
(/Users/<me>/Development/mwe/main.tex
LaTeX2e <2023-11-01> patch level 1
L3 programming layer <2024-02-20>
(/usr/local/texlive/2024/texmf-dist/tex/latex/base/article.cls
Document Class: article 2023/05/17 v1.4n Standard LaTeX document class
(/usr/local/texlive/2024/texmf-dist/tex/latex/base/size10.clo))

! LaTeX Error: File `missing.tex' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: tex)

Enter file name: 
! Emergency stop.
<read *> 
         
l.3 \input{missing}
                   ^^M
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on main.log.
pdflatex returned an error, check the log file
scons: *** [main.pdf] Error 1
```

I agree with @mwichmann that this should be treated as a system file that SCons wouldn't know how to regenerate. This would hit hard for "standard" packages distributed in TeXLive like `beamer.sty`, `standalone.sty`, etc (or any other of the myriad packages the system should have). It's analogous to trying to include a header for a library the user didn't install (You want to use libxml2? Why didn't you install it?).

If the user wanted to do something like
```tex
\documentclass{article}
\usepackage{scons}
...
```
Then they should have done something along the lines of
```python
sty = env.Command("scons.sty", ["scons.ins", "scans.idtx"], "$PDFLATEX ${SOURCE.file}", chdir=1)
```
to teach SCons how to do the Right Thing™.

I would say the main add on this PR is the (noted) incomplete test, but is that wanted?

PR #3855, opened 2021-01-05
Don't allow duplicate side effects to be added

grossag:
SCons disallows adding duplicate sources, but does not place any limits on adding duplicate side effects. This change adds code to not allow adding duplicate side effects.

I considered mimicking Node.sources_set by adding Node.side_effects_set but it seemed unnecessary and would make it harder for Environment.SideEffect to only return the list of side effects that were actually added.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -40,8 +40,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       when running multiple tests with multiple jobs.
     - Fix incorrect cache hits and/or misses when running in interactive mode by having
       SCons.Node.Node.clear() clear out all caching-related state.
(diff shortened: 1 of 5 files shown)


CHANGES.txt
```diff
@@ -40,6 +40,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       when running multiple tests with multiple jobs.
     - Fix incorrect cache hits and/or misses when running in interactive mode by having
       SCons.Node.Node.clear() clear out all caching-related state.
+    - Change Environment.SideEffect() to not add duplicate side effects. The list of returned
+      side effects will not include any duplicate side effects.
```
bdbaddog: Can you split the second sentence to a new line prefixed with NOTE:

PR #3895, opened 2021-02-27
Allow subst callables to handle default value args and functions constructed using functools.partial to set extraneous args

bdbaddog:
Allow subst callables to handle default value args and functions constructed using functools.partial to set extraneous args

This is an improved implementation of the logic in PR #3889 

These cases should now be handled
(You can find this sample SConstruct [here](https://github.com/bdbaddog/scons-bugswork/blob/main/subst_callable_class/SConstruct) )

```
import functools


class foo:
    def __init__(self, arg):
        self.arg = arg

    def __call__(self, target, source, env, for_signature):
        return self.arg + " bar"


# Will expand $BAR to "my argument bar baz"
env = Environment(FOO=foo, BAR="${FOO('my argument')} baz")

other_foo = foo(' Bananas')
env['BANANAS'] = other_foo

print("$BAR    ->%s" % env.subst("$BAR"))
print("$BANANAS->%s" % env.subst("$BANANAS"))


def afunc(target, source, env, for_signature, other_flag='one'):
    return "Fun %s" % other_flag


env['AFUNC'] = afunc
print("afunc   ->%s" % env.subst("$AFUNC"))

bfunc = functools.partial(afunc, other_flag='two')
env['BFUNC'] = bfunc
print("bfunc   ->%s" % env.subst("$BFUNC"))

def cs(target=None, source=None, env=None, for_signature=None):
    return 'cs'

env['CS'] = cs

print("CS     ->%s"%env.subst("$CS"))
```

Output:
```
$BAR    ->my argument bar baz
$BANANAS->Bananas bar
afunc   ->Fun one
bfunc   ->Fun two
CS     ->cs
scons: Building targets ...
scons: `.' is up to date.
scons: done building targets.
```

## Contributor Checklist:

### DIFF
--- a/SCons/Subst.py
+++ b/SCons/Subst.py
@@ -425,8 +425,8 @@ class StringSubber:
             # string if called on, so we make an exception in this condition for Null class
             # Also allow callables where the only non default valued args match the expected defaults
             # this should also allow functools.partial's to work.
-            if isinstance(s, SCons.Util.Null) or set([k for k, v in signature(s).parameters.items() if k in _callable_args_set or
-                                                      v.default == Parameter.empty]) == _callable_args_set:
+            if isinstance(s, SCons.Util.Null) or {k for k, v in signature(s).parameters.items() if
+                                                  k in _callable_args_set or v.default == Parameter.empty} == _callable_args_set:
 
                 s = s(target=lvars['TARGETS'],
                      source=lvars['SOURCES'],
@@ -602,9 +602,8 @@ class ListSubber(collections.UserList):
             # string if called on, so we make an exception in this condition for Null class
             # Also allow callables where the only non default valued args match the expected defaults
             # this should also allow functools.partial's to work.
-            if isinstance(s, SCons.Util.Null) or set(
-                    [k for k, v in signature(s).parameters.items() if k in _callable_args_set or
-                                                                      v.default == Parameter.empty]) == _callable_args_set:
+            if isinstance(s, SCons.Util.Null) or {k for k, v in signature(s).parameters.items() if
+                                                  k in _callable_args_set or v.default == Parameter.empty} == _callable_args_set:
 
                 s = s(target=lvars['TARGETS'],
                      source=lvars['SOURCES'],
--- a/SCons/Subst.py
+++ b/SCons/Subst.py
@@ -25,9 +25,9 @@
 
 import collections
 import re
-from inspect import signature
-import SCons.Errors
+from inspect import signature, Parameter
 
+import SCons.Errors
 from SCons.Util import is_String, is_Sequence
 
 # Indexed by the SUBST_* constants below.
@@ -328,6 +328,8 @@ def subst_dict(target, source):
     return dict
 
 
+_callable_args_set = {'target', 'source', 'env', 'for_signature'}
+
 class StringSubber:
     """A class to construct the results of a scons_subst() call.
 
@@ -335,6 +337,8 @@ class StringSubber:
     source with two methods (substitute() and expand()) that handle
     the expansion.
     """
+
+
     def __init__(self, env, mode, conv, gvars):
         self.env = env
         self.mode = mode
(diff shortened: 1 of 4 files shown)


SCons/Subst.py
```diff
@@ -593,8 +600,11 @@ def expand(self, s, lvars, within_list):
             # SCons has the unusual Null class where any __getattr__ call returns it's self, 
             # which does not work the signature module, and the Null class returns an empty
             # string if called on, so we make an exception in this condition for Null class
-            if (isinstance(s, SCons.Util.Null) or
-                set(signature(s).parameters.keys()) == set(['target', 'source', 'env', 'for_signature'])):
+            # Also allow callables where the only non default valued args match the expected defaults
+            # this should also allow functools.partial's to work.
+            if isinstance(s, SCons.Util.Null) or {k for k, v in signature(s).parameters.items() if
```
deriamis, line 605: Maybe just use a `filter` here? You're only checking and using the key, after all, and this implies creating a temporary `dict` to do the check. Using a `filter` should perform better than that.

deriamis:
This looks pretty much exactly where I was going next in #3889, apart from a comment I left you. Sorry I took a while to get back to you on it - life and work happened, as they do for me quite often. The tests look like they cover the cases I was imagining as well.

SCons/Subst.py
bdbaddog: We're using both the key and the value.. 
The key to see if the parameter name is in the list of default expected args, and the value to determine if it doesn't have a default a value.

Also items() doesn't create a dictionary, it returns a view object. So no dict should get created. See https://docs.python.org/3/library/stdtypes.html#dict-views and https://docs.python.org/3/library/stdtypes.html#typesmapping (then scroll down to items())

deriamis: Ah, no, we're talking about two different things, and I was talking about something that doesn't exist in your code. I was seeing a dictionary comprehension rather than a set comprehension on this line for some reason. I still think a `filter` might be a somewhat prettier way to do it (in my obviously biased opinion), but in the end you're doing a whole comparison anyway, so it doesn't matter performance-wise.

bdbaddog:
Ran this vs current master branch:

```
import functools
import timeit


class foo:
    def __init__(self, arg):
        self.arg = arg

    def __call__(self, target, source, env, for_signature):
        return self.arg + " bar"


# Will expand $BAR to "my argument bar baz"
env = Environment(FOO=foo, BAR="${FOO('my argument')} baz")

other_foo = foo(' Bananas')
env['BANANAS'] = other_foo

print("$BAR    ->%s" % env.subst("$BAR"))
print("$BANANAS->%s" % env.subst("$BANANAS"))


def afunc(target, source, env, for_signature, other_flag='one'):
    return "Fun %s" % other_flag


env['AFUNC'] = afunc
print("afunc   ->%s" % env.subst("$AFUNC"))

bfunc = functools.partial(afunc, other_flag='two')
env['BFUNC'] = bfunc
print("bfunc   ->%s" % env.subst("$BFUNC"))

def cs(target=None, source=None, env=None, for_signature=None):
    return 'cs'

env['CS'] = cs

print("CS     ->%s"%env.subst("$CS"))

def run_it():
    env.subst("$CS")
    env.subst("$BFUNC")
    env.subst("$AFUNC")
    env.subst("$BANANAS")
    env.subst("$BAR")

print(timeit.timeit('run_it()',globals=globals()))
```

Default iterations  1,000,000

Here's the output from new code:
```
% python ~/devel/scons/git/scons/scripts/scons.py
$BAR    ->my argument bar baz
$BANANAS->Bananas bar
afunc   ->Fun one
bfunc   ->Fun two
CS     ->cs
662.452844942
scons: Building targets ...
scons: `.' is up to date.
scons: done building targets.
```

Output from master:
```
% python ~/devel/scons/git/as_scons/scripts/scons.py
$BAR    ->my argument bar baz
$BANANAS->Bananas bar
afunc   -><function afunc at 0x105115050>
bfunc   ->functools.partial(<function afunc at 0x105115050>, other_flag='two')
CS     ->cs
687.6602115329999
scons: Building targets ...
scons: `.' is up to date.
scons: done building targets.
```

On my 2015 macbook pro
New branch: 662.452 seconds
Master        : 687.660 seconds (and not properly evaluating `afunc`, and `bfunc`)

On my ryzen 7 2700
New Branch: 284.141
Master         : 284.898 (and of course that's with subst mistakenly string-ifying the afunc and bfunc)

Ran again commenting out afunc and bfunc in the run_it() function:
On my ryzen 7 2700
New Branch: 164.403
Master         : 164.969 (and of course that's with subst mistakenly string-ifying the afunc and bfunc)  

So reasonable to assume no perf degradation against master for this.
Merging.

PR #3921, opened 2021-04-05
allow coverage to fail in travis

dmoody256:
It appears travis coverage (which runs slower in general) causes some race condition or timing issue in the test/Parallel/failed-build.py test, so the coverage job often fails.

## Contributor Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ jobs:
   allow_failures:
     - python: pypy3
     - python: 3.10-dev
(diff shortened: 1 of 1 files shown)


bdbaddog:
Declining. This test fails in regular and windows regressions as well.
Just allowing failures for this isn't a good solution.
You'd end up ignoring any other testing done in that run failing.

The best option is to just fix the test.. which thus far hasn't been simple.

PR #3922, opened 2021-04-06
Provide a custom argument escape function for `TempFileMunge` using a  new `TEMPFILEARGESCFUNC` variable

ivankravets:
## Contributor Checklist:

--------

This patch works in production (@PlatformIO) for over 3 months.


bdbaddog:
Looks good. 
Once you add a test and add the new var info to docs in `SCons/Platform/Platform.xml`  to we should be able to merge.

ivankravets:
Done! So, you can squash my commits into one. Github allows that.

SCons/Platform/PlatformTests.py
```diff
@@ -189,6 +189,36 @@ def test_TEMPFILEARGJOINBYTE(self):
...
+        env['OVERSIMPLIFIED'] = 'command'
+
+        # For tempfilemunge to operate.
+        old_actions = SCons.Action.print_actions
+        SCons.Action.print_actions = 0
+        env['TEMPFILEARGESCFUNC'] = _tempfile_arg_esc_func
+        cmd = t(None, None, env, 0)
+        # print("CMD is: %s"%cmd)
+
+        with open(cmd[-1], 'rb') as f:
```
bdbaddog, line 215: You should be able to skip opening and reading the file and instead use 
`target[0].attributes.tempfile_cmdlist[defined_cmd]`

To check for command escaping?

ivankravets: I like this approach more because we check what actually will go later to the compiler. `target[0].attributes.tempfile_cmdlist[defined_cmd]` is not the final result.

In any case, I can rewrite the test according to your advice.

bdbaddog: I'd prefer not hitting the filesystem in this case. If we're having python writing strings to file issues, it'd be more general issue. The test suite takes a while to run and any reasonable things we can do to speed it up is welcome.

So please make the change requested?
Then I think I can go ahead and merge.
Ideally we'll get a release out this weekend. (I need to review all the changes encompassed to determine version number, either 4.1.1 or 4.2.0..)

ivankravets: I'm not sure that I understand how to create a test according to your requirements. Could you give some hints? It seems you have permission to edit this PR. I would be thankful if you can correct it.

bdbaddog: If you look at the next test in that file, it checks the tempfile_cmdlist...

bdbaddog: Ah.. I am incorrect. You need to read the file for this one.   I don't seem to be able to push to your repo. Also.. You've made this change on your master. Probably best to do PR's from branches per our default github flow.

If you can make these edits to your code, then I can go ahead and merge. I've moved the default value for TEMPFILEARGESCFUNC to Defaults.py instead of in get().. 

```
% git diff platformio/master
diff --git a/SCons/Defaults.py b/SCons/Defaults.py
index 8bc0f5fb7..c108ea884 100644
--- a/SCons/Defaults.py
+++ b/SCons/Defaults.py
@@ -628,6 +628,7 @@ ConstructionEnvironment = {

     'TEMPFILE': NullCmdGenerator,
     'TEMPFILEARGJOIN': ' ',
+    'TEMPFILEARGESCFUNC': SCons.Subst.quote_spaces,
     'Dir': Variable_Method_Caller('TARGET', 'Dir'),
     'Dirs': Variable_Method_Caller('TARGET', 'Dirs'),
     'File': Variable_Method_Caller('TARGET', 'File'),
diff --git a/SCons/Platform/__init__.py b/SCons/Platform/__init__.py
index 81f3f8bd1..697a0722c 100644
--- a/SCons/Platform/__init__.py
+++ b/SCons/Platform/__init__.py
@@ -262,11 +262,11 @@ class TempFileMunge:
             prefix = '@'

         args = [
-            env.get('TEMPFILEARGESCFUNC', SCons.Subst.quote_spaces)(arg)
+            env['TEMPFILEARGESCFUNC'](arg)
             for arg in cmd[1:]
         ]
         join_char = env.get('TEMPFILEARGJOIN',' ')
-        os.write(fd, bytearray(join_char.join(args) + "\n",'utf-8'))
+        os.write(fd, bytearray(join_char.join(args) + "\n", 'utf-8'))
         os.close(fd)

         # XXX Using the SCons.Action.print_actions value directly
```

bdbaddog: Created a Pull Request against your repo with these changes.

ivankravets:
@bdbaddog everything is merged! Thanks! 🙏

PR #3969, opened 2021-07-10
Moved CacheDir permissions update code behind atomic rename operation.

dmoody256:
Moved CacheDir writeable permission code for copy to cache behind the atomic rename operation.

This fixes a potential race condition in shared cache environments where the permissions are
not writeable for a moment after the file has been renamed and other builds (users) will copy
it out of the cache.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -65,6 +65,11 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Update BuildTask to pass all targets to the progress object fixing an issue
       where multi-target build nodes only got the first target passed to the progress
       object.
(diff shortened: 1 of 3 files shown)


bdbaddog:
Can you add a blurb to CHANGES.txt and RELEASE.txt just indicating the change and purpose? (reduce chance of race condition when copying from cachedir or something like that?)

bdbaddog:
> Can you add a blurb to CHANGES.txt and RELEASE.txt just indicating the change and purpose? (reduce chance of race condition when copying from cachedir or something like that?)

Never mind. I added a blurb.

PR #3981, opened 2021-07-20
fixed ninja tool is_conftest check and allow setting disable ninja execute from env

dmoody256:
some minor update to work with mongo build before 4.2 release

## Contributor Checklist:

### DIFF
--- a/SCons/Tool/ninja/__init__.py
+++ b/SCons/Tool/ninja/__init__.py
@@ -176,8 +176,7 @@ def generate(env):
     if not NINJA_BINARY:
         raise SCons.Warnings.SConsWarning("Failed to import ninja, attempt normal SCons build.")
 
-    env["NINJA_DISABLE_AUTO_RUN"] = GetOption('disable_execute_ninja')
-
+    env["NINJA_DISABLE_AUTO_RUN"] = env.get("NINJA_DISABLE_AUTO_RUN", GetOption('disable_execute_ninja'))
     env["NINJA_FILE_NAME"] = env.get("NINJA_FILE_NAME", "build.ninja")
 
     # Add the Ninja builder.
@@ -421,9 +420,9 @@ def generate(env):
         target = self.targets[0]
         if target.get_env().get('NINJA_SKIP'):
             return
-        if target.check_attributes('ninja_file') is None or not target.is_conftest:
+        if target.check_attributes('ninja_file') is None:
             NINJA_STATE.add_build(target)
-        else:
+        else: 
             target.build()
 
     SCons.Taskmaster.Task.execute = ninja_execute
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -41,6 +41,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       Environment variables can use that rather than "$( _concat(...) $)".
     - Fix issue with exparimental ninja tool which would fail on windows or when ninja package wasn't 
       installed but --experimental=ninja was specified.
+    - As part of experimental ninja tool, allow SetOption() to set both disable_execute_ninja and
+      disable_ninja.
 
   From David H:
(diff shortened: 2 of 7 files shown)


SCons/Tool/ninja/__init__.py
```diff
@@ -419,7 +418,7 @@ def ninja_execute(self):
         target = self.targets[0]
         if target.get_env().get('NINJA_SKIP'):
             return
-        if target.check_attributes('ninja_file') is None or not target.is_conftest:
+        if target.check_attributes('ninja_file') is None and not target.is_conftest():
```
bdbaddog: I'm guessing we don't have a test which caught that this was wrong, so we should probably add one? or enhance an existing test to catch this?

SCons/Tool/ninja/__init__.py
```diff
@@ -173,8 +173,7 @@ def generate(env):
         SCons.Warnings.SConsWarning("Failed to import ninja, attempt normal SCons build.")
         return
 
-    env["NINJA_DISABLE_AUTO_RUN"] = GetOption('disable_execute_ninja')
-
+    env["NINJA_DISABLE_AUTO_RUN"] = env.get("NINJA_DISABLE_AUTO_RUN", GetOption('disable_execute_ninja'))
```
bdbaddog, line 179: Was NINJA_DISABLE_AUTO_RUN being set elsewhere pre tool initialization and then ignored? why?

SCons/Tool/ninja/__init__.py
dmoody256: okay I'll add a ninja conf test

SCons/Tool/ninja/__init__.py
dmoody256: Yeah we plan on setting it to True before loading the tool. The mongo build would prefer to set this in the SConscript then from the command line, to maintain the existing workflow.

bdbaddog: > Yeah we plan on setting it to True before loading the tool. The mongo build would prefer to set this in the SConscript then from the command line, to maintain the existing workflow.

Can't use SetOption() ?

dmoody256: it says its not settable

SCons/Tool/ninja/__init__.py
dmoody256: adding a test made me realize that the is_conftest check here is not needed. This check is inside the taskmaster.Task execute function, which for SConf it has its own execute, so ninja never sees those task executes. Its only needed in the builder execute.

The test I wrote creates a failing conftest and makes sure the ninja generation and build still proceeds, but it is dependent on is_conftest working so it will need to based off https://github.com/SCons/scons/pull/3980

SCons/Tool/ninja/__init__.py
bdbaddog: > it says its not settable

That's fixable.. I'll take a look tomorrow.
There's a fixed list of what's settable, but you should be able to modify it when the ninja tool is loaded I think. As long as your `SetOption()`  is after you initialize ninja..

dmoody256: okay but the GetOption is called during init, so if you SetOption after then it wouldn't reload that in the env var.

I can remove this change if you really want, we will just set the env var after the tool is loaded.

SCons/Tool/ninja/__init__.py
```diff
@@ -421,9 +420,9 @@ def ninja_execute(self):
         target = self.targets[0]
         if target.get_env().get('NINJA_SKIP'):
             return
-        if target.check_attributes('ninja_file') is None or not target.is_conftest:
```
dmoody256, line 424: The test I added revealed this was not needed because the SConf has its own isolated Taskmaster and task

PR #3997, opened 2021-08-06
Expanded ninja Mkdir to also support Mkdir actions.

dmoody256:
Mkdir command was not support directly in an Command builder, so I added some stuff to improve support.

## Contributor Checklist:


bdbaddog:
So what's the purpose here?

You had implemented for a action function named `MkdirFunc` but not for `Mkdir`, so you added logic to handle too?

Also minor update on the way the mkdir command string is generated?

SCons/Tool/ninja/__init__.py
```diff
@@ -186,7 +186,7 @@ def generate(env):
     env.Append(BUILDERS={"Ninja": ninja_builder_obj})
 
     env["NINJA_ALIAS_NAME"] = env.get("NINJA_ALIAS_NAME", "generate-ninja")
-    env['NINJA_DIR'] = env.get("NINJA_DIR", env.Dir(".ninja").path)
+    env['NINJA_DIR'] = env.get("NINJA_DIR", ".ninja")
```
dmoody256: making it a dir causes scons to create a node which then attempts to create the directory for ninja, but ninja will already do this, so it ends up with an unnecessary warning message that a directory already exists.

SCons/Tool/ninja/NinjaState.py
```diff
@@ -536,6 +536,7 @@ def __init__(self, env):
             # TODO: use command action #3573
             "installFunc": _install_action_function,
             "MkdirFunc": _mkdir_action_function,
+            "Mkdir": _mkdir_action_function,
```
dmoody256, line 552: while investing the warning message about a directory already existing I dug into ninja's conversion of Mkdir function, and found that the alternate use via Command(action='Mkdir') was not supported so I added it.

SCons/Tool/ninja/Rules.py
```diff
@@ -37,16 +37,16 @@ def _install_action_function(_env, node):
...
         # used by anything that depends on these but commonly this is
         # hit with a node that will depend on all of the fake
         # srcnode's that SCons will never give us a rule for leading
         # to an invalid ninja file.
         "variables": {
             # On Windows mkdir "-p" is always on
-            "cmd": "{mkdir}".format(
-                mkdir="mkdir $out & exit 0" if env["PLATFORM"] == "win32" else "mkdir -p $out",
+            "cmd": "mkdir {args}".format(
+                args = ' '.join(get_outputs(node)) + " & exit /b 0" if env["PLATFORM"] == "win32" else "-p " + ' '.join(get_outputs(node)),
```
dmoody256, line 49: the windows bit wasn't working here, because a CMD rule is supposed to a be a singular command, in which the $in and $out are supplied after the command. If you want to have something more complicated, then GENERATED_CMD is more suited, which lets you handle setting the inputs and outputs.

dmoody256:
> So what's the purpose here?
> 
> You had implemented for a action function named `MkdirFunc` but not for `Mkdir`, so you added logic to handle too?
> 
> Also minor update on the way the mkdir command string is generated?

There were several things around the mkdir and ninja support that were not working correctly. I added comments to each of them in the PR.

bdbaddog:
@dmoody256 
The new added tests passes against current upstream.
So it must not be testing the changes.
Do you have a simple repro for each of the issues? Dir creation and missing MkDir?

dmoody256:
> @dmoody256
> The new added tests passes against current upstream.
> So it must not be testing the changes.
> Do you have a simple repro for each of the issues? Dir creation and missing MkDir?

@bdbaddog I looked into it and it does pass current master. The reason is that the ninja tool is making scons callbacks for the MkDir actions it does not know how to perform itself. The changes in these PR make sure ninja knows how to correctly create directories by itself without needing to call back to scons. I can update the test to ensure that no callbacks are performed when executed.

SCons/Tool/ninja/__init__.py
bdbaddog: Should this instead be `#/.ninja` ?

test/ninja/mkdir_function_command.py
```diff
@@ -0,0 +1,123 @@
...
+    ninja.__file__,
+    os.pardir,
+    'data',
+    'bin',
+    'ninja' + _exe))
+
+test.write('SConstruct', """
+SetOption('experimental','ninja')
+DefaultEnvironment(tools=[])
+env = Environment()
```
bdbaddog, line 50: You should be ok to add `tool=[]` in the`Environment()` here as well, since you only need to have ninja tool installed right?
(It does impact the speed of tests, especially on windows where it will try to initialize MSVC which is a slow process)

test/ninja/mkdir_function_command.py
```diff
@@ -0,0 +1,123 @@
...
+env.Tool('ninja')
+env.Dir('test0')
```
bdbaddog, line 52: What's the purpose of these `Dir(..)`'s?

bdbaddog: You have Dir()'s as sources?

test/ninja/mkdir_function_command.py
```diff
@@ -0,0 +1,123 @@
...
+os.rmdir('test0')
+os.rmdir('test1/test2')
+os.rmdir('test1')
+os.rmdir('test3/test4')
+os.rmdir('test3')
+os.rmdir('test5')
+os.rmdir('test6/test7')
+os.rmdir('test6')
+
+test.must_not_exist([
```
bdbaddog, line 86: You're verifying that os.rmdir()'s working here?

test/ninja/mkdir_function_command.py
dmoody256: Dir will cause a Mkdir function to be invoked, and this is make sure that it is translated to ninja either if SCons use `Dir` or `Command(..., Mkdir)`

test/ninja/mkdir_function_command.py
dmoody256: I suppose it is a little silly. I was verifying that the work dir is in a clean state, because the proceeding tests expect that.

PR #4034, opened 2021-10-18
Fix crash when a signal handler is set from outside Python

bquistorff:
If a signal handler is set from outside python (e.g., C extension) then retrieving the signal returns None ([docs](https://docs.python.org/3/library/signal.html#signal.getsignal)), but that value can't be then used to set the signal. In this case reset to default handler after we're done.

I ran into this problem running SCons from a Python environment created inside Stata (a statistical program). It could conceivably happen in other scenarios if a C extension registers a signal handler.

## Contributor Checklist:

I do not know how to setup a unit test for this issue. Nor did a see an appropriate place to update the documents.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -57,6 +57,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Change SCons.Scanner.Base to ScannerBase. Old name kept as an alias
       but is now unused in SCons itself.
 
+  From Brian Quistorff:
+    - Fix crash when scons is run from a python environement where a signal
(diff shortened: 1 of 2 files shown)


bdbaddog:
Thanks for the PR.
Please address the sider complaints.
Also please move the conditional logic outside of the function calls.

I find it much harder to read and maintain.

bdbaddog:
How would we create a test for this?

bquistorff:
Existing tests should ensure this change doesn't affect normal setups. On my setup this change fixed the crash, but I don't know how to write a simple unit test for it.

PR #4140, opened 2022-04-20
[Ninja] make sure ninja always executes when run from SCons

dmoody256:
This was brought up by a user on the mailing list:
https://pairlist4.pair.net/pipermail/scons-users/2022-April/008882.html

The issue was the user made no changes to the build and reran scons with ninja enabled. SCons did not rebuild the ninja file unless the ninja file was named as an explicit target.

## Contributor Checklist:


SCons/Tool/ninja/NinjaState.py
```diff
@@ -529,52 +572,71 @@ def generate(self):
...
+
+            if pidfile:
+                with open(pidfile) as f:
+                    pid = int(f.readline())
+                    try:
+                        os.kill(pid, signal.SIGINT)
+                    except OSError:
+                        pass
+
+                    # wait for the server process to fully killed
```
dmoody256: I would rather use psutil to do this, but that means another dependency, so this is a way to do it manually. Not sure it handles every edge case.

SCons/Tool/ninja/NinjaState.py
```diff
@@ -77,17 +80,36 @@ def __init__(self, env, ninja_file, ninja_syntax):
...
+        # The daemon port should be the same across runs, unless explicitly set
+        # or if the portfile is deleted. This ensures the ninja file is deterministic
+        # across regen's if nothings changed. The construction var should take preference,
+        # then portfile is next, and then otherwise create a new random port to persist in
+        # use.
+        scons_daemon_port = None
+        os.makedirs(get_path(self.env.get("NINJA_DIR")), exist_ok=True)
+        scons_daemon_port_file = str(pathlib.Path(get_path(self.env.get("NINJA_DIR"))) / "scons_daemon_portfile")
+
+        if env.get('NINJA_SCONS_DAEMON_PORT') is not None:
```
bdbaddog: This would probably be clearer as 

```
if x:
elif y:
elif z:
```

SCons/Tool/ninja/NinjaState.py
```diff
@@ -209,7 +232,7 @@ def __init__(self, env, ninja_file, ninja_syntax):
                 "restat": 1,
             },
             "TEMPLATE": {
-                "command": f"{sys.executable} {pathlib.Path(__file__).parent / 'ninja_daemon_build.py'} {scons_daemon_port} {get_path(env.get('NINJA_DIR'))} $out",
+                "command": f"{sys.executable} {pathlib.Path(__file__).parent / 'ninja_daemon_build.py'} $PORT {env.get('NINJA_DIR').abspath} $out",
```
bdbaddog: would it be worth changing `{env.get('NINJA_DIR').abspath}` to $ABS_NINJA_DIR and define it up in self.variables?

bdbaddog:
Could you switch to using OrderedDict() for self.rules, etc? Then you can skip sorting at the end?

SCons/Tool/ninja/__init__.py
```diff
@@ -65,8 +65,11 @@ def ninja_builder(env, target, source):
     print("Generating:", str(target[0]))
 
     generated_build_ninja = target[0].get_abspath()
-    NINJA_STATE.generate()
-
+    try:
+        NINJA_STATE.generate()
+    except Exception as e:
+        print(f"ERROR: an excetion occurred while generating the ninja file:\n{traceback.format_exc()}")
```
bdbaddog: Should probably cause this to propagate to SCons's exception handling?

dmoody256:
> Could you switch to using OrderedDict() for self.rules, etc? Then you can skip sorting at the end?

We would still need to sort because sorting is there to prevent unnecessary regens. The OrderedDict is just to guarantee that the sorted order is not lost from a standard dictionary. The OrderedDict is not necessary in python3.7 where dict is guaranteed to maintain order.

SCons/Tool/ninja/__init__.py
dmoody256: yep, updated

SCons/Tool/ninja/NinjaState.py
dmoody256: yeah I also did several others to clean it up a bit

SCons/Tool/ninja/NinjaState.py
dmoody256: yep, cleaned up the logic

SCons/Tool/ninja/Utils.py
```diff
@@ -258,6 +259,19 @@ def ninja_noop(*_args, **_kwargs):
     """
     return None
 
+def ninja_recursive_sorted_dict(build):
```
bdbaddog: Add a docstring here?

CHANGES.txt
```diff
@@ -63,6 +63,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Ninja: Fix issue where Configure files weren't being properly processed when build run
       via ninja.
     - Added ninja mingw support and improved ninja CommandGeneratorAction support.
+    - Made ninja tool force the ninja file as the only target, and make it only update the ninja
```
bdbaddog: Seems like this PR is actually 2 PR's?
1 . Fix build.ninja file to always be generated the same (given same build)
2. Do some magic with default targets (which I'm not sure I fully understand yet).

Is that right?

dmoody256: Yeah the determinism could be separated

dmoody256: @bdbaddog okay this PR now has determinism changes removed.

SCons/Util.py
```diff
@@ -2123,6 +2124,40 @@ def print_time():
...
+
+def wait_for_process_to_die(pid):
+    """
+    Wait for specified process to die, or alternatively kill it
+    NOTE: This function operates best with psutil pypi package
+    """
+    # wait for the process to fully killed
+    try:
+        import psutil
+        while True:
```
dmoody256, line 2137: Just want to note this loop could be troublesome if for some reason the process wont die. Could be a default time out and related error message to save future headaches

bdbaddog: @dmoody256 - what seems like a reasonable timeout? probably as a parameter to this.. maybe 5 seconds?

doc/user/external.xml
```diff
@@ -377,6 +399,12 @@ conda install -c conda-forge ninja
             implement those actions via shell commands in the &ninja; build file.
         </para>
 
+        <para>
+            When ninja runs the generated &ninja; build file, ninja will launch &scons; as a daemon and feed commands
+            to that &scons; process which &ninja; is unable to build directly.  This daemon will stay alive until
+            explicitly killed or it times out. The timeout is set by &cv-link-NINJA_SCONS_DAEMON_KEEP_ALIVE; .
```
dmoody256: Maybe add a sentence here: "The daemon will be restarted if the SConscripts change or the build changes in a way that ninja determines it needs to regenerate the build.ninja file"

bdbaddog: Done.

PR #4169, opened 2022-06-07
[NINJA] Added option to allow scons to determine if it should skip ninja regeneration.

dmoody256:
## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -135,8 +135,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       Note that these are called for every build command run by SCons. It could have considerable
       performance impact if not used carefully.
       to connect to the server during start up.
-    - Ninja: added option to skip ninja regeneration if scons can determine the ninja file does
-      not need to be regenerated.
+    - Ninja: added option "--skip-ninja-regen" to enable skipping regeneration of the ninja file
+      if scons can determine the ninja file doesnot need to be regenerated, which will also
+      skip restarting the scons daemon. Note this option is could result in incorrect rebuilds
+      if scons Glob or scons generated files are used in ninja build target's command lines.
     - Ninja: Added new alias "shutdown-ninja-scons-daemon" to allow ninja to shutdown the daemon.
       Also added cleanup to test framework to kill ninja scons daemons and clean ip daemon logs.
       NOTE: Test for this requires python psutil module. It will be skipped if not present.
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -116,6 +116,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       Note that these are called for every build command run by SCons. It could have considerable
       performance impact if not used carefully.
       to connect to the server during start up.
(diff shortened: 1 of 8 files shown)


CHANGES.txt
```diff
@@ -135,6 +135,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       Note that these are called for every build command run by SCons. It could have considerable
       performance impact if not used carefully.
       to connect to the server during start up.
+    - Ninja: added option to skip ninja regeneration if scons can determine the ninja file does
```
bdbaddog: What is the option? How do you enable? (disable)

RELEASE.txt
```diff
@@ -107,6 +107,8 @@ CHANGED/ENHANCED EXISTING FUNCTIONALITY
         require delayed expansion to be enabled which is currently not supported and is
         typically not enabled by default on the host system. The batch files may also require
         environment variables that are not included by default in the msvc environment.
+- Ninja: added option to skip ninja regeneration if scons can determine the ninja file does
```
bdbaddog: same as CHANGES.txt comment

SCons/Tool/ninja/__init__.py
```diff
@@ -66,8 +66,12 @@ def ninja_builder(env, target, source):
     print("Generating:", str(target[0]))
 
     generated_build_ninja = target[0].get_abspath()
-    NINJA_STATE.generate()
-
+    try:
+        NINJA_STATE.generate()
+    except Exception:
+        raise SCons.Errors.BuildError(
+            errstr=f"ERROR: an excetion occurred while generating the ninja file:\n{traceback.format_exc()}",
```
bdbaddog: Typo: exception

test/ninja/ninja_test_sconscripts/sconstruct_ninja_determinism
```diff
@@ -0,0 +1,20 @@
+import SCons
+import random
+SetOption('experimental','ninja')
+SetOption('skip_ninja_regen', True)
```
bdbaddog, line 4: If I comment this line out, the test still passes? Is that expected?

dmoody256: yeah actually this flag is not relevant to the current test. This PR makes ninja files deterministic, but the flag is for controlling skipping the regeneration and restarting of the daemon. The flag is to protect users who may be using Glob or SCons generated files in command lines. Its assumed if the user enables the flag, they know what they are doing.

I will update this test to also make sure the mtime of the ninja file did not change, which would indicate it was never overwritten.

dmoody256:
@bdbaddog fixed the notes, and updated the test

test/ninja/ninja_test_sconscripts/sconstruct_ninja_determinism
bdbaddog: o.k. so run this test with master branch code and it should fail then?

dmoody256: yeah I tested that in the latest

PR #4222, opened 2022-08-25
Implement new parallel scheduler

acmorrow:
There is a lot more to do before this could merge, but I wanted to get the idea out there for discussion. Things I know still need to be done.

- A top level `--scheduler=blah` switch or something similar to allow choosing between different schedulers
- Performance test results showing how it fares vs the existing parallel scheduler.
- Probably putting each scheduler in its own file
- Style nits, formatting, etc.
- Fixing tests.
- Changes files, etc.

But I don't want to spend a ton of time on that stuff unless there is a consensus around pursuing this idea.

Overall, the approach here is that instead of having one thread that drives the taskmaster and N worker threads that just execute jobs, there are just N worker threads on an equal footing. Any single worker thread can drive the taskmaster while other threads execute jobs. Once a thread has driven the taskmaster through to finding a job that requires execution, that thread goes off to execute the task and allows the next free worker thread to pick up the responsibility for turning the taskmaster crank and processing any pending results.

The nice thing about this model is that it doesn't require the system to first find N executable jobs (or exhaust that search) before it ever retires a single execution, like the current `Parallel` scheduler does. That model just doesn't work well when you are building with a very high number of jobs (say, 500) against a distributed compile cluster like icecream. The existing `Parallel` spends a ton of upfront time walking odd parts of the graph in an effort to find those N jobs because it never sees any completions. It results in odd bursty / sawtooth behavior. Attempts to build a better heuristic for when to issue vs when to retire didn't really seem promising. The free threaded approach here doesn't need any heuristic. Whichever thread is doing the serial phase w.r.t. the taskmaster is responsible for both retiring executed tasks in order to unblock graph exporation and then finding new work.

I've attempted to add enough comments that it should be possible to get a pretty good sense of the flow of control by reading through it, but I'm happy to answer any questions.

I probably won't be able to get back to this for a few weeks, and as noted above it is very much work in progress, so I'm making this a draft PR.

### DIFF
--- a/SCons/Taskmaster/Job.py
+++ b/SCons/Taskmaster/Job.py
@@ -145,7 +145,7 @@ class Jobs:
             else:
                 os._exit(2)  # pylint: disable=protected-access
 
-        self.old_sigint  = signal.signal(signal.SIGINT, handler)
+        self.old_sigint = signal.signal(signal.SIGINT, handler)
         self.old_sigterm = signal.signal(signal.SIGTERM, handler)
         try:
             self.old_sighup = signal.signal(signal.SIGHUP, handler)
@@ -169,6 +169,7 @@ class Jobs:
             signal.signal(signal.SIGHUP, sigterm_to_use)
         except AttributeError:
             pass
+
 
 class Serial:
     """This class is used to execute tasks in series, and is more efficient
@@ -211,7 +212,7 @@ class Serial:
                     try:
                         raise SCons.Errors.BuildError(
                             task.targets[0], errstr=interrupt_msg)
-                    except:
+                    except Exception:
                         task.exception_set()
                 else:
                     task.exception_set()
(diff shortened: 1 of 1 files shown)


acmorrow:
@bdbaddog - I've made this into a real PR. With @dmoody256's perf results from discord I think it is ready for closer review. Things to go over:

- What should we do with the `print` statements as far as turning them into opt-in tracing?
- Do we want to put this under a flag?
- What should we do about some tests that don't pass. They maybe aren't very useful tests? I'm not sure.
- Does the `Serial` class still offer anything in a multicore world? Should we just always use my `NewParallel` for all `-j` values?

acmorrow:
I rebased, squashed, and force pushed this PR. The historical version is still available here: https://github.com/acmorrow/scons/tree/new-parallel-backup

PR #4248, opened 2022-10-21
cxx20 module support for gcc

loonycyborg:
## C++ module support

This PR is aimed is aimed at adding support for C++20 modules for gcc, perhaps also adding infrastructure that could be reused by clang/msvc. Intended interface is to enable module support only if CXXMODULEPATH is set to non-empty path and store module files there.

## Contributor Checklist:


loonycyborg:
GCC general documentation on modules: 
https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Modules.html
Module mapper documentation:
https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Module-Mapper.html

mwichmann:
Since there are some other links in the long-running discussion, plus notes on how clang does things, thought I'd mention the Bazel project's issue bazelbuild/bazel#4005

ZhenyuC-Dev:
Hey how's this commit going? Any one still following up?
If not, I'm kind of interested in this and could be able to pick it up.

mwichmann:
Any contributions here welcome.  We perhaps should have a better place for "discussion" about the right design for this (the one in the PR may in fact be that).  In the meantime, this link can be a partial starting point, since it describes a problem and has more links:

https://github.com/urnathan/libcody

There's a newer version of 1184:

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1184r2.pdf

Not sure if the mapper/build communication stuff is overkill for a first cut at supporting modules. As I said - maybe a separate location for discussion?

ZhenyuC-Dev:
Good to know. But sorry I'm kind of confused by now. Cuz I saw this commit in bazel's issues. But it seems that this is another repo. I'm wondering what's the relationship between these two?

mwichmann:
relationship? just a problem that all the build systems have to solve, in the face of a not really unified compiler approach to tackling it.

loonycyborg:
We have a channel for this on Discord. 
I'll likely prioritize working on this if some other people actually review/try it.

SCons/Tool/gxx.py
```diff
@@ -67,6 +224,17 @@ def generate(env) -> None:
     env['CCDEPFLAGS'] = '-MMD -MF ${TARGET}.d'
     env["NINJA_DEPFILE_PARSE_FORMAT"] = 'gcc'
 
+    env['__CXXMODULEINIT__'] = init_mapper
+    if sys.platform != "win32":
+        env['__GCCMAPPERMODE__'] = "domain_socket"
+        env['__GCCMAPPERFLAGS__'] = "=$CXXMODULEPATH/socket"
```
jasoncwik, line 230: Remove the extra `=` here. It ends up failing:

```
/usr/include/c++/13/map: error: failed opening mapper '=cxx-modules/socket'
```

mwichmann: yeah, looks like a typo

loonycyborg: https://gcc.gnu.org/onlinedocs/gcc/gcc-command-options/c%2B%2B-modules.html#module-mapper <- `=` introduces a socket based mapper.

jasoncwik: Wow, you're right. Weird syntax :)

Looking more closely it was failing because the mountpoint didn't allow a domain socket to be created.

loonycyborg: Are you using some unusual os?
FTR I usually run it like this: 
```
~/vcs-checkouts/scons/test/CXX/CXX-modules-fixture [cxx20-gcc|✚ 1…7] $ python ../../../scripts/scons.py toolset=g++
scons: Building targets ...
g++ -o test.o -c -fmodules-ts -fmodule-mapper==cxx-scons-modules/socket?test.cpp -std=c++20 test.cpp
g++ -o itest.o -c -fmodules-ts -fmodule-mapper==cxx-scons-modules/socket?itest.cpp -std=c++20 itest.cpp
g++ -x c++-system-header -c -fmodules-ts -fmodule-mapper==cxx-scons-modules/socket?@system-header/iostream -std=c++20 iostream
g++ -o main.o -c -fmodules-ts -fmodule-mapper==cxx-scons-modules/socket?main.cpp -std=c++20 main.cpp
g++ -o scons-module-test itest.o test.o main.o
scons: done building targets.
```

jasoncwik: Yeah, it's a build container with a bind mount from a Mac. Just had to move the module directory to /var.

loonycyborg:
Full blown compiler/buildsystem communication is only used because it's the only way to see actual paths for system headers to support header imports of form `import <iostream>;`. So mostly I used it because I desire this feature for my own use. However I prefer `import std;` and thus wouldn't have particular need for system header imports anymore once this is implemented:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106852
Maybe could redo this to use plain file lists instead of mapper then.

PR #4253, opened 2022-10-27
Change --taskmaster trace logging to use Python's logging module.  Added logging to NewParalllel Job class

bdbaddog:
Current Implementation requires setting
`USE_NEW_PARALLEL` shell variable to some true value when you call SCons
This is temporary and will be changed to an experimental feature in next PR.

## Contributor Checklist:


SCons/Taskmaster/Job.py
```diff
@@ -547,24 +567,28 @@ def _work(self):
                     # in the condvar. Some other thread will come back
                     # here with a completed task.
                     if self.state == ExperimentalParallel.State.STALLED and completed_task:
-                        # print(f"XXX {threading.get_ident()} Detected stall with completed task, bypassing wait")
+                        if self.trace:
+                            self.trace_message(f"XXX {threading.get_ident()} Detected stall with completed task, bypassing wait")
```
acmorrow: I think you should replace the `XXX`'s with `ExperimentalParallel: ` or similar in these messages.

bdbaddog: With the latest updates you'll see something like this in trace.log

```
Job.NewParallel._work(): 6180548608 Gained exclusive access
Job.NewParallel._work(): 6180548608 Starting search
```
(or LegacyParallel for the "current" ParallelJob)

PR #4456, opened 2024-01-02
Fix sphinx's conf.py to handle SCons versions like 4.6.0.post1

bdbaddog:
Fix sphinx's conf.py to handle SCons versions like 4.6.0.post1

## Contributor Checklist:


mwichmann (commented):
I wouldn't actually include the patch version, just recognize and toss it. by definition, nothing changed so it doesn't need calling out.

bdbaddog:
we don't require non "public" APIs to stay unchanged between patch releases.. so maybe it's worth keeping?

PR #4467, opened 2024-01-30
Parallel cache push strawman

acmorrow:
## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -65,6 +65,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       Python to 3.6, and it was not until Python 3.7 where `threading` became
       default supported. In practice, we expect most real world Python 3.6 deployments
       will have `threading` support enabled, so this will not be an issue.
+    - CacheDir writes no longer happen within the taskmaster critical section,
+      and therefore can run in parallel with both other CacheDir writes and the
+      taskmaster DAG walk.
 
   From Mats Wichmann:
     - Add support for Python 3.13 (as of alpha 2). So far only affects
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -63,6 +63,8 @@ IMPROVEMENTS
   the new scheduler is now used for -j1 builds as well.
   NOTE: This should significantly improve SCons performance for larger parallel builds
         (Larger -j values)
+- CacheDir writes no longer happen within the taskmaster critical section, and therefore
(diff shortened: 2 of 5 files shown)


acmorrow:
@bdbaddog - This seems to pass all tests, but please let me know if you would prefer to do this under a `--experimental=???` flag and I can rework it like that.

bdbaddog:
One concern/question.
Previously the node had `.cached` which you could use to see if the file was cached or not.
Now there's only a check to see if it shouldn't be cached, but no way to tell if the node has actually been cached?

Would restoring .cached break anything in this change?

acmorrow:
> One concern/question. Previously the node had `.cached` which you could use to see if the file was cached or not. Now there's only a check to see if it shouldn't be cached, but no way to tell if the node has actually been cached?
> 
> Would restoring .cached break anything in this change?

I don't think it would break anything to restore it. I took it out mostly because as far as I could see it was only being used to transfer info about caching from `execute` to `executed`, and we don't do that anymore. And I think the test that examined it will still need to go away. So it really will be more or less unused. Fine with me either way though.

bdbaddog:
> > One concern/question. Previously the node had `.cached` which you could use to see if the file was cached or not. Now there's only a check to see if it shouldn't be cached, but no way to tell if the node has actually been cached?
> > Would restoring .cached break anything in this change?
> 
> I don't think it would break anything to restore it. I took it out mostly because as far as I could see it was only being used to transfer info about caching from `execute` to `executed`, and we don't do that anymore. And I think the test that examined it will still need to go away. So it really will be more or less unused. Fine with me either way though.

I think it's worth adding back in. 
While not part of the published API, I can see someone useing it, and we may find it useful going forward with remote caching,etc..

acmorrow:
> > > One concern/question. Previously the node had `.cached` which you could use to see if the file was cached or not. Now there's only a check to see if it shouldn't be cached, but no way to tell if the node has actually been cached?
> > > Would restoring .cached break anything in this change?
> > 
> > 
> > I don't think it would break anything to restore it. I took it out mostly because as far as I could see it was only being used to transfer info about caching from `execute` to `executed`, and we don't do that anymore. And I think the test that examined it will still need to go away. So it really will be more or less unused. Fine with me either way though.
> 
> I think it's worth adding back in. While not part of the published API, I can see someone useing it, and we may find it useful going forward with remote caching,etc..

@bdbaddog - done.

PR #4606, opened 2024-09-29
Fixed inconsistent usage of keep_local and local_only kwarg in Help() function, its implementation and docs

hedger:
## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -49,6 +49,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       cl.exe-compatible command line switches.
     - Some manpage cleanup for the gettext and pdf/ps builders.
     - Some clarifications in the User Guide "Environments" chapter.
+    - Fixed wrong naming in the documentation of "keep_local" parameter for 
+      the Help() method.
 
   From Alex James:
     - On Darwin, PermissionErrors are now handled while trying to access
--- a/SCons/Script/Main.py
+++ b/SCons/Script/Main.py
@@ -601,8 +601,8 @@ def ValidateOptions(throw_exception: bool = False) -> None:
     OptionsParser.preserve_unknown_options = False
     OptionsParser.parse_args(OptionsParser.largs, OptionsParser.values)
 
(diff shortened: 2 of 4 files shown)


bdbaddog:
Not sure why you believe this change is needed?
local_only, means only the local addoptions args are in the help.. what's the concern here?

hedger:
@bdbaddog the concern is that documentation is inconsistent with actual behavior. 

Currently, [the docs](https://scons.org/doc/4.8.1/HTML/scons-man.html#f-Help) mention the following prototype for the Help() function:

> `env.Help(text, append=False, local_only=False)`

While actual implementation in https://github.com/SCons/scons/blob/4f3e3026bc455b3b47666567f273b2f84f932d32/SCons/Script/__init__.py#L254 uses the `keep_local` kwarg name.

I made the documentation consistent with the actual implementation - in order to not break any third-party code that relies on existing behavior. I also renamed the kwarg inside the implementation to avoid confusion.

bdbaddog:
Seems like it would be simpler to just make everything `local_only` and not `keep_local`

also calling `Help()` in your SConscript calls SCons.SConscript.Help(), and not PrintHelp() (not directly).
Help() calls HelpFunction() which calls PrintHelp()

So the public interface is SCons.SConscript.Help(), and that's what's misdocumented.
(and mis doc-string'ed too I think)

Thoughts @mwichmann ?

CHANGES.txt
```diff
@@ -49,6 +49,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       cl.exe-compatible command line switches.
     - Some manpage cleanup for the gettext and pdf/ps builders.
     - Some clarifications in the User Guide "Environments" chapter.
+    - Fixed wrong naming in the documentation of "keep_local" parameter for
```
bdbaddog, line 52: This should be changed to be attributed to you and not to mwichmann.
Please add a section with your name here ( it should be alphabetical by last name)

bdbaddog:
Replacing this with #4670

PR #4634, opened 2024-11-12
User Guide: Add mention of `no_progress` option when user passes -Q

ghost:
Accordingly to [changes to 4.0.0](https://scons.org/scons-400-is-available.html)

## Contributor Checklist:

### DIFF
--- a/doc/user/command-line.xml
+++ b/doc/user/command-line.xml
@@ -557,6 +557,11 @@ foo.in
       </row>
 
       <row>
+      <entry><literal>no_progress</literal></entry>
+      <entry><option>-Q</option></entry>
+      </row>
+
+      <row>
       <entry><literal>no_site_dir</literal></entry>
       <entry><option>--no-site-dir</option></entry>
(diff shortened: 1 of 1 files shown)


mwichmann:
Sigh... that's the problem with maintaining similar information in multiple places.  At a glance, it looks like other rows are missing from the table in the User Guide compared to the manpage.

mwichmann:
30 rows in the user guide table, 41 in the manpage `GetOption` table, 19 in the manpage `SetOption` table (of course, not all the options are "settable"). 

@bdbaddog - is it better to point to the `GetOption/SetOption` entries here than try to duplicate?

mwichmann:
Should we update this to say something like "Get/Set Option table out of date compared to manpgage"?

bdbaddog:
@mwichmann - is there anyway we can single source this table? Stick in a file and include both places?

mwichmann:
not that easy with current setup, though possible, I think.

PR #4799, opened 2025-12-14
Fix debug include printing for multi-source

mwichmann:
Rather obscure bug - if `--debug=includes` is selected, the analysis stopped after the first source file of a given target. For the common case of C/C++ projects, the object builders are single-source, so the problem would not be noticed.

Adds a distinct testcase, because the main debug-includes test uses the C compiler, which won't trigger this.

No doc impacts, although the documentation of the option could perhaps be a little more explicit.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -35,6 +35,7 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       a warning for years, but now it's a fatal error).
     - Improve covarage of API doc build by ignoring any setting of
       __all__ in a package and not showing inherited members from optparse.
+    - Fix --debug=includes for case of multiple source files.
 
 
 RELEASE 4.10.1 - Sun, 16 Nov 2025 10:51:57 -0700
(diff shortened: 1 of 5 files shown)


mwichmann:
Just so it's mentioned here, do the printouts make sense? For source file `f2.in`, we get this display: 
```
+-f2.in
  +-a.inc
  | +-[a.inc]
  | +-b.inc
  |   +-[a.inc]
  |   +-[b.inc]
  +-[b.inc]
```
Why all the extra noise?  `a.inc` doesn't depend on `b.inc` nor on itself, etc. The standard tree display doesn't do this:
```
+-.
  +-SConstruct
  +-a.inc
  +-b.inc
  +-f.out
  | +-f1.in
  | +-f2.in
  | +-a.inc
  | +-b.inc
  +-f1.in
  +-f2.in
```

mwichmann:
hmmm, wonder if this is due to the Fake scanner... may need to be a little bit more clever on that.

PR #4805, opened 2025-12-26
Update CONTRIBUTING for editors and for ai

mwichmann:
Now that the vi/emacs stubs at the end of every file have been removed, add a short note about possibly needing to setup your editor. .editorconfig is better anyway, it works on far more editors than just vim and emacs, but not every situation will work out of the box.

Also add a section on AI use, and fix a few links.

This change is purely to "project files" , no changes to SCons itself.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -21,7 +21,7 @@ to hop onto the `SCons Discord Server <https://discord.gg/bXVpWAy>`_
 to chat.  You don't have to use the Discord app,
 as there is a web interface (does require an account).
 You can also use the
-`SCons Users Mailing List <https://pairlist4.pair.net/mailman/listinfo/scons-users>`_.
+`SCons Users Mailing List <https://mail.python.org/mailman3/lists/scons-users.python.org/>`_.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -30,10 +30,10 @@ Here are some important pointers to other resources:
 
   * `SCons Home Page <https://scons.org>`_
   * `Github Project Page <https://github.com/scons/scons>`_
-  * `Bugs and Feature Requests <https://scons.org/bugs.html>`_
+  * `Bugs and Feature Requests <https://scons.org/bugs>`_
```
bdbaddog: pretty sure this should be bugs.html

bdbaddog: fixed and pushed to this branch.

mwichmann: either works... github recognizes them without the `.html` (i.e. it wasn't wrong before)

bdbaddog: using without html yields.. 
`Not Found
The requested URL was not found on this server.`

PR #4813, opened 2026-01-06
Updates for mailing list change

mwichmann:
The recent change of mailing list host, which changed both the address to send mail to and the administrative URL, is reflected in the codebase - mainly to the various README files, but also a number of one-off references.

This is only visible in the SCons code error messages emitted if a deprecated Python version is used, and in the metadata if an rpm package is generated - neither are "functional" changes.

### DIFF
--- a/README-SF.rst
+++ b/README-SF.rst
@@ -41,7 +41,7 @@ Latest Version
 Before going further, you can check that the package you have is the latest
 version at the SCons download page:
 
-        http://www.scons.org/pages/download.html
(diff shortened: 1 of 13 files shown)


jcbrill:
@mwichmann Possible typo: "installating" on line 128 of README.rst.

mwichmann:
> @mwichmann Possible typo: "installating" on line 128 of README.rst.

Yup, thanks. Fixed.

PR #4810, opened 2026-01-02
Use scons as the default MSVSSCONS variable

madmiraal:
Fixes #4809

Currently, the default value for `MSVSSCONS` used to populate the `vcxproj` file's `NMake*CommandLine` lines, uses a rounadabout way to launch SCons. It launches python with `-c` and a string for the following python script:
```
from os.path import join
import sys
sys.path = [
    join(sys.prefix, 'Lib', 'site-packages', 'scons-4.9.1'),
    join(sys.prefix, 'scons-4.9.1'),
    join(sys.prefix, 'Lib', 'site-packages', 'scons'),
    join(sys.prefix, 'scons')
] + sys.path;
import SCons.Script
SCons.Script.main()
```
If the script can't find the `scons` module in the modified `sys.path` it fails.

It should be safe to assume that `scons` is (or at least should be) in the user's path; as it's needed to run the `SConstruct` that is used to create the `vcxproj` file. However, if `scons` is not in the user's path, the user is free to modify `MSVSSCONS` variable to include the full path to their `scons`.

Therefore, this PR sets the default `MSVSSCONS` variable to `scons`, and removes the scaffolding used to create the previous default value.

## Contributor Checklist:

* [ ] I have created a new test or updated the unit tests to cover the new/changed functionality.
* [ ] I have updated `CHANGES.txt` and `RELEASE.txt` (and read the `README.rst`).
* [ ] I have updated the appropriate documentation

Issue #4809, opened 2026-01-02
The vcxproj file created using MSVSProject fails to build the project when SCons installed via scoop


madmiraal:
**Describe the bug**
The `vcxproj` file created using `MSVSProject` fails to build the project with the following error:
```
1>Traceback (most recent call last):
1>  File "<string>", line 1, in <module>
1>ModuleNotFoundError: No module named 'SCons'
1>C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.MakeFile.Targets(45,5): error MSB3073: The command "echo Starting SCons && "C:\Users\Marcel\scoop\apps\python\current\python.exe" -c "from os.path import join; import sys; sys.path = [ join(sys.prefix, 'Lib', 'site-packages', 'scons-4.9.1'), join(sys.prefix, 'scons-4.9.1'), join(sys.prefix, 'Lib', 'site-packages', 'scons'), join(sys.prefix, 'scons') ] + sys.path; import SCons.Script; SCons.Script.main()" -C "C:\Users\Marcel\Temp\SCons" -f SConstruct "C:\Users\Marcel\Temp\SCons\hello.exe"" exited with code 1.
1>Done building project "Hello.vcxproj" -- FAILED.
```

**Required information**
* Link to SCons Users thread discussing your issue: https://discord.com/channels/571796279483564041/1456022630631997662
* Version of SCons: 4.9.1
* Version of Python: 3.10.1
* Which python distribution if applicable (python.org, cygwin, anaconda, macports, brew,etc): scoop (https://scoop.sh/)
* How you installed SCons: scoop
* What Platform are you on? (Linux/Windows and which version): Windows 11 Home 24H2
* How to reproduce your issue?  Please include a small self contained reproducer. Likely a SConstruct should do for most issues.
```
env = Environment()
sources = ['hello.cpp']
includes = ['hello.h']
program = env.Program(target='hello.exe', source=sources)
env.MSVSProject(
    target='Hello' + env['MSVSPROJECTSUFFIX'],
    srcs=sources,
    incs=includes,
    buildtarget=program,
    variant='Release',
)
```
* How you invoke scons (The command line you're using "scons --flags some_arguments"): `scons`

I think the problem lies with the way MSVSProject creates the NMake*Commands in the vcxproj file:
```
[...]

PR #4817, opened 2026-01-18
Fixed GH Issue 4809. scoop.sh installed scons (and likely scons-local )created msvs project files would fail.

bdbaddog:
Fixed GH Issue #4809 4809. scoop.sh installed scons (and likely scons-local )created msvs project files would fail.

See #4810 for discussion.
This is the implementation of the fix discussed in that PR.

Largely the work of.
@jcbrill  & @madmiraal

Note: I tested this live with the SConstruct from #4809 and [scoop.sh](https://scoop.sh) installed scons
(And then locally editing scoop installed scons' msvs.py"
 
## Contributor Checklist:

### DIFF
--- a/SCons/Script/Main.py
+++ b/SCons/Script/Main.py
@@ -1478,11 +1478,14 @@ def _exec_main(parser, values) -> None:
         _main(parser)
 
 
-def main() -> None:
+def main(script_path) -> None:
     global OptionsParser
     global exit_status
     global first_command_start
     global ENABLE_JSON
+    global SCONS_SCRIPT_PATH
+
+    SCONS_SCRIPT_PATH = script_path
 
     # Check up front for a Python version we do not support.  We
     # delay the check for deprecated Python versions until later,
--- a/testing/framework/TestSConsMSVS.py
+++ b/testing/framework/TestSConsMSVS.py
@@ -900,9 +900,10 @@ print("self._msvs_versions =%%s"%%str(SCons.Tool.MSCommon.query_versions(env=Non
             project_guid = PROJECT_GUID
 
         if 'SCONS_LIB_DIR' in os.environ:
-            exec_script_main = f"from os.path import join; import sys; sys.path = [ r'{os.environ['SCONS_LIB_DIR']}' ] + sys.path; import SCons.Script; SCons.Script.main()"
+            exec_script_main = f"import sys; sys.path = [ r'{os.environ['SCONS_LIB_DIR']}' ] + sys.path; import SCons.Script; SCons.Script.main()"
         else:
-            exec_script_main = f"from os.path import join; import sys; sys.path = [ join(sys.prefix, 'Lib', 'site-packages', 'scons-{self.scons_version}'), join(sys.prefix, 'scons-{self.scons_version}'), join(sys.prefix, 'Lib', 'site-packages', 'scons'), join(sys.prefix, 'scons') ] + sys.path; import SCons.Script; SCons.Script.main()"
+            exec_script_main = f"import sys; sys.path = [ scons_parent ] + sys.path; import SCons.Script; SCons.Script.main()"
+
         exec_script_main_xml = exec_script_main.replace("'", "&apos;")
 
         result = input.replace(r'<WORKPATH>', workpath)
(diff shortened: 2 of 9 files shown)


testing/framework/TestSConsMSVS.py
```diff
@@ -900,9 +900,10 @@ def msvs_substitute(
             project_guid = PROJECT_GUID
 
         if 'SCONS_LIB_DIR' in os.environ:
-            exec_script_main = f"from os.path import join; import sys; sys.path = [ r'{os.environ['SCONS_LIB_DIR']}' ] + sys.path; import SCons.Script; SCons.Script.main()"
+            exec_script_main = f"import sys; sys.path = [ r'{os.environ['SCONS_LIB_DIR']}' ] + sys.path; import SCons.Script; SCons.Script.main()"
         else:
-            exec_script_main = f"from os.path import join; import sys; sys.path = [ join(sys.prefix, 'Lib', 'site-packages', 'scons-{self.scons_version}'), join(sys.prefix, 'scons-{self.scons_version}'), join(sys.prefix, 'Lib', 'site-packages', 'scons'), join(sys.prefix, 'scons') ] + sys.path; import SCons.Script; SCons.Script.main()"
+            exec_script_main = f"import sys; sys.path = [ scons_parent ] + sys.path; import SCons.Script; SCons.Script.main()"
+
```
jcbrill, line 907: There's a bug in the else clause: `[ scons_parent ]` should be an f-string with either a preceding `scons_parent` variable definition or an inline path calculation similar to the construction in `SCons/Tool/msvs.py`.

By default, the else clause is not exercised by the test suite.

Was reviewing the code when making #4818 functionally equivalent (with other changes).

bdbaddog: comment no longer applies as code has been replaced.

SCons/Script/Main.py
```diff
@@ -1478,11 +1478,14 @@ def lookupmodule(self, filename: str) -> str | None:
         _main(parser)
 
 
-def main() -> None:
+def main(script_path) -> None:
```
jcbrill, line 1481: Bugs of omission?

`SCons/SCript/Main` main function now takes a required `script_path` argument.

Missing:
* `SCons/__main__.py` needs to pass a `script_path` argument.
* The `bin/scons-test.py` function `do_nothing` needs to accept a `script_path` argument.

I could be wrong.

bdbaddog: Good catch. Updating.

bdbaddog: How'd you find those? Are we missing some tests which should catch such? (python -m scons ?)

jcbrill: I just searched the source code for combinations of `Script.main` (e.g. regex `Script\.main\(` most likely).

Potential issues:
* *The change from `main()` to `main(script_path)` can/will break user code*.
  
  Basically anyone doing the same thing: `import SCons.Script; SCons.Script.main()`.

* Isn't `main` the entry point when building the wheel wrapper executables for Windows?
  ```
  [console_scripts]
  scons = SCons.Script.Main:main
  scons-configure-cache = SCons.Utilities.ConfigureCache:main
  sconsign = SCons.Utilities.sconsign:main
  ```
  I thought that the entry point was not supposed to take arguments but I certainly could be wrong.  A required argument could be problematic.
* `scons.py` is not the only way SCons will be called.
  
  For example, the zipapp (e.g., `scons-local-4.10.1.pyz`) will be called via `__main__.py`. 

  However, adding `os.path.abspath(__file__)` will not work for the zippapp.  An invocation like `S:\SCons\test-4809\scons-local-4.10.1.pyz\__main__.py` will fail.

  Something like `os.path.dirname(os.path.abspath(__file__))` would be needed to use the name of the zipapp file itself.  An invocation like `S:\SCons\test-4809\scons-local-4.10.1.pyz` should work.

* Changing the *default* generated msvs script code path from using `PYTHON_ROOT` and `PYTHONPATH` *which has allowed deferred evaluation of the location of python and SCons since version 1.0* to a hard-coded absolute path could break some user code.

  Both `PYTHON_ROOT` (or something similar) and `PYTHONPATH` were/are useful when using python and SCons from a USB stick and/or when physically moving folders between computers as it avoids hard-coded absolute paths.

  Edit: This is a different animal when actually launching the VS GUI which can be done via batch file(s) via the SCons project root to open the generated .sln in the appropriate build folder.

Don't shoot the messenger.

bdbaddog: Not shooting the messenger, just wanted to see what I should have done to catch those.

PR #4822, opened 2026-01-28
Make sure consistently using one docstring style

mwichmann:
The remaining files that have old-style aka reST-style docstrings are switched to Google style.  This does *not* remove the inline reST-style annotations for functions, classes, modules, etc. - that's a different topic, we want to keep those so API docs are generated with hyperlinks to the referenced element.

In a bit of "burying the lede"... there was an actual bugfix here, in `Defaults.py` a copy-pasted chunk ended up with a reference to a variable that was not in scope at that point (in `processDefines`). Was glancing at `Defaults.py` in an IDE to see what it thought was broken, and it was red on that one.

Some minor de-linting in `Defaults.py` and `Node/FS.py`, and fix an annotation error in `FS.py` (`invalidate_node_memos`) - of course that one has no runtime effect.

This has no doc or test changes, it's nearly all docstrings, type annotations and moving imports - none of which have runtime effects. 

## Contributor Checklist:

### DIFF
--- a/SCons/Defaults.py
+++ b/SCons/Defaults.py
@@ -45,6 +45,7 @@ import SCons.Builder
 import SCons.CacheDir
 import SCons.Environment
 import SCons.Errors
+import SCons.Node.FS
 import SCons.PathList
 import SCons.Scanner.Dir
 import SCons.Subst
@@ -606,7 +607,7 @@ def processDefines(defs) -> list[str]:
             # TODO: do we need to quote value if it contains space?
             dlist.append(f"{name}={value[0]}")
         else:
-            dlist.append(str(define[0]))
+            dlist.append(str(defs[0]))
     elif is_Dict(defs):
         for macro, value in defs.items():
             if value is None:
@@ -685,10 +686,13 @@ class Variable_Method_Caller:
 def __libversionflags(env, version_var, flags_var):
     """
     if version_var is not empty, returns env[flags_var], otherwise returns None
-    :param env:
-    :param version_var:
-    :param flags_var:
-    :return:
+
+    Args:
+        env:
+        version_var:
+        flags_var:
+
+    Returns:
     """
     try:
         if env.subst('$' + version_var):
@@ -701,11 +705,14 @@ def __libversionflags(env, version_var, flags_var):
 def __lib_either_version_flag(env, version_var1, version_var2, flags_var):
     """
     if $version_var1 or $version_var2 is not empty, returns env[flags_var], otherwise returns None
-    :param env:
-    :param version_var1:
-    :param version_var2:
-    :param flags_var:
-    :return:
+
+    Args:
+        env:
+        version_var1:
+        version_var2:
+        flags_var:
+
+    Returns:
     """
     try:
         if env.subst('$' + version_var1) or env.subst('$' + version_var2):
--- a/testing/framework/TestCmd.py
+++ b/testing/framework/TestCmd.py
@@ -534,7 +534,7 @@ def pass_test(self=None, condition: bool = True, function=None) -> None:
     a condition argument is supplied; if so the completion processing
     takes place only if the condition is true.
 
-    the test passes only if the condition is true.
+    The test passes only if the condition is true.
 
     Args:
         self: a test class instance. Must be passed in explicitly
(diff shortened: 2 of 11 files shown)


testing/framework/TestCmd.py
```diff
@@ -551,16 +551,20 @@ def pass_test(self=None, condition: bool = True, function=None) -> None:
     sys.exit(0)
 
 
-def match_exact(lines=None, matches=None, newline=os.sep):
+def match_exact(
+    lines: str | list | None = None,
+    matches: str | list | None = None,
```
Repiteo:
```suggestion
    lines: str | list[str] | None = None,
    matches: str | list[str] | None = None,
```

testing/framework/TestCmd.py
```diff
@@ -577,19 +581,21 @@ def match_exact(lines=None, matches=None, newline=os.sep):
     return 1
 
 
-def match_caseinsensitive(lines=None, matches=None):
+def match_caseinsensitive(
+    lines: str | list | None = None, matches: str | list | None = None
```
Repiteo:
```suggestion
    lines: str | list[str] | None = None, matches: str | list[str] | None = None
```

testing/framework/TestCmd.py
```diff
@@ -603,15 +609,15 @@ def match_caseinsensitive(lines=None, matches=None):
     return 1
 
 
-def match_re(lines=None, res=None):
+def match_re(lines: str | list | None = None, res: str | list | None = None) -> int | None:
```
Repiteo:
```suggestion
def match_re(
    lines: str | list[str] | None = None, res: str | list[str] | None = None
) -> int | None:
```

testing/framework/TestCmd.py
```diff
@@ -635,18 +641,20 @@ def match_re(lines=None, res=None):
     return 1
 
 
-def match_re_dotall(lines=None, res=None):
+def match_re_dotall(
+    lines: str | list | None = None, res: str | list | None = None
```
Repiteo:
```suggestion
    lines: str | list[str] | None = None, res: str | list[str] | None = None
```

testing/framework/TestCmd.py
```diff
@@ -2152,21 +2160,19 @@ def do_chmod(fname) -> None:
                     do_chmod(os.path.join(dirpath, name))
             do_chmod(top)
 
-    def write(self, file, content, mode: str = 'wb'):
+    def write(self, file: str | list, content: str | bytes, mode: str = 'wb'):
```
Repiteo:
```suggestion
    def write(self, file: str | list[str], content: str | bytes, mode: str = 'wb'):
```

Repiteo (commented):
The returning `None on failure, 1 on success.` makes my heart ache, but I understand this is a non-functional pass. However, seeing as you *are* updating the type hints, the container types should be explicit

mwichmann:
> The returning `None on failure, 1 on success.` makes my heart ache

Indeed.

mwichmann:
These look fine to me, @bdbaddog okay for the changes to be added?

testing/framework/TestCmd.py
mwichmann: Same problem as for `match_exact`

testing/framework/TestCmd.py
mwichmann: Hmmm, this isn't going to work as annotated anyway... if `lines` is the default `None`, then:
```py
    if not is_List(lines):
        lines = lines.split(newline)
```
is going to throw an `AttributeError` cause `None` has no `split` method. The same for `matches`, for that matter. I'm thinking a default of an empty string is better?

testing/framework/TestCmd.py
mwichmann: Same problem as for `match_exact`

bdbaddog:
> These look fine to me, @bdbaddog okay for the changes to be added?

My type hints fu is weak. If they look good you, go ahead..

mwichmann:
So.. should we eliminate the `None` type option and `None` default for the observed and expected arguments to the four match functions?  As mentioned above, they will raise an exception if `None` actually made it through as an argument value, nor do they make any actual sense - how can you call a function to compare two strings or lists of string (or regex strings) if one or both aren't even supplied?

SCons/Defaults.py
```diff
@@ -606,7 +607,7 @@ def processDefines(defs) -> list[str]:
             # TODO: do we need to quote value if it contains space?
             dlist.append(f"{name}={value[0]}")
         else:
-            dlist.append(str(define[0]))
+            dlist.append(str(defs[0]))
```
bdbaddog, line 610: Is this change correct?

mwichmann: Yeah. This is a stanza duplicated from a few lines above (about 583) when I did the revamp. Except in the earlier stanza it's inside a `for define in defs:` loop and here it isn't, so `define` isn't valid. Could have just used `name`, I guess, since that's `defs[0]` from the unpack in 605.

mwichmann: This is alluded to in the PR intro above, by the way.

PR #4823, opened 2026-01-31
Simplify and speedup compilation database generation

Repiteo:
Gives the compilation database code some TLC, as the previous implementation caused a lot of `--debug=explain` output bloat. While I think this will always be an expensive action to *some* extent, here's an anecdotal comparison of the difference I saw from an MSVC build of the Godot Engine that's already been built:

|Method|Node Count|Build Time|
|-|-|-|
|Disabled|17,161|00:16.61|
|Enabled (Old)|20,008 (+2,847)|00:19.54 (+2.93)|
|Enabled (New)|17,163 (+2)|00:18.38 (+1.77)|

The one area I couldn't figure out was making the entries function as a dependency for the compilation database itself, so it's still set to always build.

## Contributor Checklist:


SCons/Tool/compilation_db.py
```diff
@@ -33,163 +33,65 @@
...
 """
 
-import json
-import itertools
 import fnmatch
-import SCons
+import itertools
+import json
 
+import SCons
```
mwichmann: Not a comment on the PR specifically since this is an existing line: we've generally tended to be more explicit about imports. `import SCons` obviously works, though an IDE doesn't really like it, so we should swing back someday and do the imports that are actually used (Default, Action, Builder, Scanner, etc.)

bdbaddog:
Can you give a high level summary of your changes? (But lower than made it faster.. ;)

SCons/Tool/compilation_db.py
Repiteo: Went ahead and adjusted this to account for that, as it makes sense to tackle this in the same pass

Repiteo:
The most obvious change is the removal of `__CompilationDbNode`, as that ended up being an unnecessary middleman. I think the original intent was to have those somehow function as dependencies for the compilation database at MongoDB, but they were set to always build since their introduction in this repo, so they were doing extra work for no real benefit. I'm not certain how much *faster* that specific change ended up being, but it was **the** reason that the overall amount of nodes decreased by nearly 3k in my test scenario

A secondary boost was from only having to create the commandstr actions once-per-variable. The old implementation was effectively doing a lambda call each time it was validating a commandstr, instead of falling back on a cached function.

Tertiary changes include making the output file always use Unix eol and UTF-8 encoding, as well as using more modern syntax to make some sections less verbose. Also removed the scanner because it did nothing. It quite literally did nothing. It might've at one point in MongoDB, but it was totally vestigial

bdbaddog:
I'm a bit slammed at the moment and this will take a deeper dive to review, so please be patient for feedback..

bdbaddog:
What happens if your build system has more than one Environment()? Seems like with the global gone, it would only output the information for items built with the current Environment()?

Repiteo:
Hmm, that's tricky. On one hand, it'd be convenient to allow multiple compiledb outputs based on the environment that called it. On the other hand, that's very much NOT how the current implementation handles it, and I don't want this PR changing observed functionality

I'll go ahead and bring back the global for this PR

SCons/Tool/compilation_db.py
```diff
@@ -202,75 +112,67 @@ def compilation_db_emitter(target, source, env):
...
-            function=scan_compilation_db, node_class=None
+            ASPPSuffixes, [(static_obj, GEN_ASPPCOM), (shared_obj, GEN_ASPPCOM)]
         ),
+    ):
+        emitter_old = builder.emitter.get(suffix)
+        # Only setup emitters for Tools supported by the environment.
+        if emitter_old:
+            builder.emitter[suffix] = ListEmitter(env.Flatten(emitter_old) + [emitter])
+
+    env["BUILDERS"]["CompilationDatabase"] = Builder(
```
bdbaddog, line 162: Any reason not to use this instead of the if x not in env form?

env["COMPILATIONDB_USE_ABSPATH"] = env.get("COMPILATIONDB_USE_ABSPATH",False)

Repiteo: Oh yeah, that's much simpler!

SCons/Tool/compilation_db.py
```diff
@@ -54,142 +55,51 @@
...
-
-    target[0].write(entry)
-
-
 def write_compilation_db(target, source, env) -> None:
-    entries = []
-
-    use_abspath = env['COMPILATIONDB_USE_ABSPATH'] in [True, 1, 'True', 'true']
-    use_path_filter = env.subst('$COMPILATIONDB_PATH_FILTER')
+    DIRECTORY = env.Dir("#").get_abspath()
```
bdbaddog: Style question, why use all caps?

bdbaddog:
Other than style question, it should be ready to merge.

SCons/Tool/compilation_db.py
Repiteo: To indicate that they're constant values; that's about it

bdbaddog: Given their within a function, please switch them to just be lower case.
We're not upper casing any other such variables in any other logic in the code base, or at least that I'm aware of.

Repiteo: Done!

PR #4824, opened 2026-02-02
Implement type hints for Environment and environment utilities

Repiteo:
Followup to Node type hinting PRs

Hints for Environment are particularly odd, as a lot of cases would technically want to derive from `SubstitutionEnvironment` instead of `Environment.Base`. However, I made the type hints act as if `Environment.Base` **is** the base class. This is because, in practice, absolutely nobody is using `SubstitutionEnvironment`, and it might even be refactored away in the future(?). The files adjusted were:
- `SCons/Environment.py`
- `SCons/Node/__init__.py` (`Environment` → `EnvironmentBase`)
- `SCons/Node/FS.py` (`Environment` → `EnvironmentBase`)
- `SCons/Script/SConscript.py` (Specifically `SConsEnvironment` and related functions)
- `SCons/Util/envs.py` (Utility scripts for environments)

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -26,6 +26,7 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
     - Purge vim/emac local variable bloat.
     - Implement type hints for Node subclasses.
     - Ruff: Handle F401 exclusions more granularly, remove per-file exclusions.
+    - Implement type hints for Environment and environment utilities.
 
   From William Deegan:
     - Fix SCons Docbook schema to work with lxml > 5
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -98,6 +98,7 @@ DEVELOPMENT
 - Unify internal "_null" sentinel usage.
 - Docbook tests: improve skip message, more clearly indicate which test
   need actual installed system programs (add -live suffix).
+- Implement type hints for Environment and environment utilities.
 
 - MSVC: Added a host/target batch file configuration table for Visual
   Studio 2026.  Visual Studio 2026 removed support for 32-bit arm targets.
--- a/SCons/Environment.py
+++ b/SCons/Environment.py
@@ -2403,7 +2403,7 @@ class Base(SubstitutionEnvironment):
             executor.add_post_action(action)
         return nodes
 
-    def Alias(self, target: str, source: str | Node | list[str | Node] = [], action=None, **kw) -> list[Node]:
+    def Alias(self, target: str | Node | list[str | Node], source: str | Node | list[str | Node] = [], action=None, **kw) -> list[Node]:
         """Create an alias *target* that depends on *source*."""
         tlist = self.arg2nodes(target, self.ans.Alias)
         if not is_List(source):
@@ -2570,12 +2570,12 @@ class Base(SubstitutionEnvironment):
         return tlist
 
     @overload
-    def Dir(self, name: str, *args, **kw) -> DirNode: ...
+    def Dir(self, name: str | Node, *args, **kw) -> DirNode: ...
 
     @overload
-    def Dir(self, name: list[str], *args, **kw) -> list[DirNode]: ...
+    def Dir(self, name: list[str | Node], *args, **kw) -> list[DirNode]: ...
 
-    def Dir(self, name: str | list[str], *args, **kw) -> DirNode | list[DirNode]:
+    def Dir(self, name: str | Node | list[str | Node], *args, **kw) -> DirNode | list[DirNode]:
         """Create Dir node(s) for *name*."""
         s = self.subst(name)
         if is_Sequence(s):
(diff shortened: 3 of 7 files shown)


mwichmann:
Tackling a big one, I see :-)  This will take a bit to review - and it seems to have blown up a few tests.  In `EnvironmentTests`, there seem to be two tests (`test_ParseConfig` and `test_ParseFlags`) for tuples which need to be preserved as such (when they're the key, value pair of a cpp define), but which have flipped to lists, and one test (`test_AppendENVPath`) which hasn't kept the expected ordering. Not sure how typing changes can cause that. `UtilTests` also has some unexpected ordering changes.

Repiteo:
Hmm, I might've left a functional change in there while testing return types; it'd explain the tests acting up anyway. Will take a look later

bdbaddog:
I asked Gemini for feedback on the type annotations, here's what it came up with. Thoughts?

The changes in this PR introduce extensive type annotations across the SCons codebase, particularly in the core environment and node management modules. After
  reviewing the changes in SCons/Environment.py, SCons/Util/envs.py, SCons/Node/FS.py, SCons/Node/__init__.py, and SCons/Script/SConscript.py, I have identified
  several areas where the type annotations could be improved or corrected.


  Incorrect or Overly Restrictive Type Annotations


   1. `SCons/Environment.py` - `Alias` method:
      The target parameter is annotated as str, but SCons aliases can also be Node objects or lists of strings/Nodes.
       * Current: target: str
       * Corrected: target: str | Node | list[str | Node]


   2. `SCons/Environment.py` - `Dir`, `File`, and `Entry` methods:
      The name parameter is annotated as str | list[str]. However, these methods frequently accept Node objects (e.g., passing a File node to env.File()).
       * Current: name: str | list[str]
       * Corrected: name: str | Node | list[str | Node] (or more specific Node types like FileNode for File)


   3. `SCons/Environment.py` - `VariantDir` method:
      The variant_dir and src_dir parameters are annotated as str, but the implementation explicitly handles DirNode objects.
       * Current: variant_dir: str, src_dir: str
       * Corrected: variant_dir: str | DirNode, src_dir: str | DirNode


   4. `SCons/Node/FS.py` - `invalidate_node_memos` function:
      The targets parameter is annotated as str | list[Base | str]. This prevents passing a single Base node directly, even though the implementation supports it.
       * Current: targets: str | list[Base | str]
       * Corrected: Base | str | list[Base | str]


   5. `SCons/Environment.py` - `NoSubstitutionProxy` methods:
      The subst and subst_target_source methods are annotated with string: str. However, the base Environment.subst method handles list[str]. If a list is passed
  to the proxy, it will return the list, but the type hint will be incorrect.
       * Current: string: str
       * Corrected: string: str | list[str]

  Observations and Fixes


   * `EnvironmentBase` Alias: The use of EnvironmentBase = Base at the end of SCons/Environment.py is a good pattern to handle circular dependencies and potential
     overrides in type checking.


  Overall, the type annotations are a significant improvement, but adjusting the restrictive hints mentioned above would better reflect SCons's flexible API and
  prevent unnecessary static analysis errors.

mwichmann:
That's weird... the comments in the email show some backward-looking comments in "Observations and Fixes", which has four bullets,,and three of those don't seem to be in the comment in the PR itself, which has one bullet ???? Makes me nervous...

bdbaddog:
> That's weird... the comments in the email show some backward-looking comments in "Observations and Fixes", which has four bullets,,and three of those don't seem to be in the comment in the PR itself, which has one bullet ???? Makes me nervous...

I think it diffed the branch vs master and since this branch branched before some of the fixes in master presently it analyzed more changes than are in this branch.  I removed 3 of the 4 observations and fixed section as they referred to changes in master..

mwichmann:
> > That's weird... the comments in the email show some backward-looking comments in "Observations and Fixes", which has four bullets,,and three of those don't seem to be in the comment in the PR itself, which has one bullet ???? Makes me nervous...
> 
> I think it diffed the branch vs master and since this branch branched before some of the fixes in master presently it analyzed more changes than are in this branch. I removed 3 of the 4 observations and fixed section as they referred to changes in master..

okay, that would explain it. thanks.

bdbaddog:
@Repiteo - can you take a look at the Gemini sugguests and make those changes if they seem sane and/or let me know if it's ok for me to?

mwichmann:
Seems like we should have a name for the frequently seen `str | Node | list[str | Node]` pattern....

mwichmann:
The numbered observations seem about right to me.

bdbaddog:
> Seems like we should have a name for the frequently seen `str | Node | list[str | Node]` pattern....

Add that to sctypes ?

mwichmann:
> > Seems like we should have a name for the frequently seen `str | Node | list[str | Node]` pattern....
> 
> Add that to sctypes ?

maybe... but on the other hand, it's a type only for type checking. We had a `SCons.Util.sctyping` for a while but it's been factored back out.

Repiteo:
@bdbaddog Those are fair observations! While passing nodes/lists in many of these contexts is unorthodox and not really documented, tests indeed exists for basically every case mentioned so it's fair to integrate them

@mwichmann iirc, it was removed because the abstraction it provided was able to be substituted in the files themselves by taking advantage of `TYPE_CHECKING`. Alias shorthands were never part of it originally, and isn't a niche that `TYPE_CHECKING` would replicate without bloat. I don't think we'd *need* to bring back `sctyping`—`sctypes` could fulfill the same purpose—but if we wanted a file *exclusively* for type hint logic, then I can see merit in that. Not a blocker for this PR in any case, but I'll test a 3.9 environment to see how much typing-specific logic could get centralized

PR #4825, opened 2026-02-03
Remove Python 3.7 support; deprecate Python 3.8 support

Repiteo:
In setting up #4824, I've realized just how limited some of the type-aware tools are with a 3.7 baseline. While that alone wouldn't be enough to want to push for a version bump, I was curious and checked if any other areas of the repo are waiting on a version bump from 3.7 specifically. As it turns out, there were multiple! Combined with 3.7 no longer being supported on several GitHub runners (`macos-15` & `ubuntu-24.04`), it made sense to drop that version outright & migrate over the functionality waiting on a 3.8 basline.

At this time, this PR is split into two commits; the second is dedicated to deprecating Python 3.8. While not as pertinent as dropping 3.7, even 3.8 has been well past eol for years now, and some of our analysis tools (mypy) have a minimum version of 3.9, so it's a reasonable target for an upcoming baseline. With that said, if this is too forward or warrants an entirely separate PR/discussion, then I can easily remove that commit and keep everything focused on the 3.7 removal instead.

## Contributor Checklist:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,17 +35,17 @@ environment:
   SCONS_CACHE_MSVC_CONFIG: "true"
   matrix:
     # Test oldest and newest supported Pythons, and a subset in between.
-    # Skipping 3.8, 3.10, 3.12 at this time
+    # Skipping 3.10, 3.12 at this time
     - WINPYTHON: "Python313"
     - WINPYTHON: "Python311"
     - WINPYTHON: "Python39"
-    - WINPYTHON: "Python37"
+    - WINPYTHON: "Python38"
 
 # remove sets of build jobs based on criteria below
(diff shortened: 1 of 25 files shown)


mwichmann:
This was in the plan, though the discussion we had earlier was to deprecate both 3.7 and 3.8 and not directly drop 3.7 without a cooling off period.

bdbaddog:
Can you split this PR in two, one with logic to deprecate 3.7 and 3.8, and then another with changes enabled by that?

bdbaddog:
@Repiteo  - Is there anything here which isn't covered in #4826  or #4827?
Can we close this?

Repiteo:
Those two cover everything; closing!

PR #4826, opened 2026-02-03
Deprecate Python 3.7 & 3.8 support

Repiteo:
Originally part of #4825, this PR simply deprecates Python 3.7 & 3.8. The only other change is removing a check in `ActionTests.py`, as I realized that was conditional on 3.6 specifically, so it being bumped to 3.7 was a mistake

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,7 +8,7 @@ NOTE: The 4.0.0 release of SCons dropped Python 2.7 support. Use 3.1.2 if
   Python 2.7 support is required (but note old SCons releases are unsupported).
 NOTE: Since SCons 4.3.0, Python 3.6.0 or above is required.
 NOTE: Since SCons 4.9.0, Python 3.7.0 or above is required.
-
+The last release to support Python 3.7 or Python 3.8 is NEXT_RELEASE.
 
 RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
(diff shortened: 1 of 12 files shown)


mwichmann:
The deprecated 3.7 failed a test (`SCONSFLAGS.py`(. Side effect of us being so fussy about exact outputs - the (expected) presence of a deprecation message shouldn't blow it up...

bdbaddog:
> The deprecated 3.7 failed a test (`SCONSFLAGS.py`(. Side effect of us being so fussy about exact outputs - the (expected) presence of a deprecation message shouldn't blow it up...

Didn't we handle this in the testing framework? (or is my memory failing me on this one?)

mwichmann:
There's something for the deprecation itself. There's not for an individual test that gets extra output due to using a deprecated Python version.  I'm kind of surprised it hit only one test. We should be able to come up with something - it might *possibly* be related to the change to runtest to add logging.

bdbaddog:
> There's something for the deprecation itself. There's not for an individual test that gets extra output due to using a deprecated Python version. I'm kind of surprised it hit only one test. We should be able to come up with something - it might _possibly_ be related to the change to runtest to add logging.

Looks like the failing part is some additional checks @mwichmann added in 2024, and we've not had any deprecations since then so it didn't get hit til now.. 
Taking a look at a fix now.

mwichmann:
I don't have the energy to try to PR a PR again, so just attaching a patch to `test/SCONSFLAGS.py` here. It wasn't really checking the right thing for the two "unknown option" cases - it needed to be looking for the deprecation message (if any) to be present there. Did some cleanup while at it.

[SCONSFLAGS.patch](https://github.com/user-attachments/files/25133599/SCONSFLAGS.patch)

PR #4827, opened 2026-02-03
Remove Python 3.7 & 3.8 support

Repiteo:
Originally part of #4825 and building off of #4826, this PR aims to integrate the changes made possible by dropping Python 3.7 & 3.8 support. Marking as a draft until 3.7 & 3.8 are properly designated as "deprecated".

## Contributor Checklist:

### DIFF
--- a/README-local
+++ b/README-local
@@ -43,7 +43,7 @@ scons-local package, or any SCons package, at the SCons download page:
 EXECUTION REQUIREMENTS
 ======================
 
-Running SCons requires Python 3.7 or higher. There should be no other
+Running SCons requires Python 3.9 or higher. There should be no other
 dependencies or requirements to run standard SCons.
 
 The last release to support Python 3.7 or Python 3.8 is 4.11.0.
--- a/SCons/Util/hashes.py
+++ b/SCons/Util/hashes.py
@@ -330,14 +330,8 @@ def hash_file_signature(fname: str, chunksize: int=65536, hash_format=None) -> s
     """
     m = _get_hash_object(hash_format)
     with open(fname, "rb") as f:
-        while True:
-            blck = f.read(chunksize)
-            if not blck:
-                break
-            m.update(to_bytes(blck))
-        # TODO: can use this when base is Python 3.8+
-        # while (blk := f.read(chunksize)) != b'':
-        #     m.update(to_bytes(blk))
+        while (blk := f.read(chunksize)):
+            m.update(to_bytes(blk))
 
     return m.hexdigest()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = ["setuptools"]
 [project]
 name = "SCons"
 description = "Open Source next-generation build tool."
-requires-python = ">=3.7"
+requires-python = ">=3.9"
 license = "MIT"  # PEP 639 form (new - setuptools >= 77.0)
 # Should include docbook license, but this fails:
 # license = "MIT AND DocBook-stylesheet"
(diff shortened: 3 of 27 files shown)


README-local
```diff
@@ -43,9 +43,10 @@ scons-local package, or any SCons package, at the SCons download page:
 EXECUTION REQUIREMENTS
 ======================
 
-Running SCons requires Python 3.7 or higher. There should be no other
+Running SCons requires Python 3.8 or higher. There should be no other
 dependencies or requirements to run standard SCons.
 
+The last release to support Python 3.7 was NEXT_RELEASE.
```
mwichmann: Need another way to express this - we don't want to turn it on processing into "the release that dropped 3.7" when it should be "the release prior to the one that dropped 3.7".

bdbaddog: Understood, but I think in this case our intent is next release 4.10.2 with above message, then following release 4.11.0 and obsolete.. so It'll work for now?  LAST_37_RELEASE ? LAST_38_RELEASE ?

mwichmann: I think we just have to keep track of that one and fix it manually as appropriate. I'm not saying we need to invent a new mechanism for something that applies to only one feature: a comment about a removed Python. Also, I think we haven't decided whether to "flip the switch" (removal) in the next release or not. Whatever the next release is shouldn't be 4.10.2 anyway, we're beyond "minor bugfix release" at this point.

Repiteo: Gonna not even bother with trying to update the documentation/changelog at this time; it gave a ton of conflicts when trying a standard merge commit and just made a ton of noise that would need to be micro-managed. I'll worry about those files once their `NEXT_RELEASE` is resolved

pyproject.toml
```diff
@@ -97,7 +95,7 @@ dist-dir = "build/dist"
 dist-dir = "build/dist"
 
 [tool.ruff]
-target-version = "py37" # Lowest python version supported
+target-version = "py38" # Lowest python version supported
```
mwichmann: No suggestion to do anything now while 4.11 is still pending, but just to record: this should have been py39.

SCons/Util/hashes.py
```diff
@@ -330,14 +330,8 @@ def hash_file_signature(fname: str, chunksize: int=65536, hash_format=None) -> s
...
-        while True:
-            blck = f.read(chunksize)
-            if not blck:
-                break
-            m.update(to_bytes(blck))
-        # TODO: can use this when base is Python 3.8+
-        # while (blk := f.read(chunksize)) != b'':
-        #     m.update(to_bytes(blk))
+        while (blk := f.read(chunksize)):
+            m.update(to_bytes(blk))
```
mwichmann, line 334: Just to record something here for when this is revisited: it would be nice to untangle the messy dual approach for pre-3.9 and 3.9+ versions, since 3.9 will be the base. There's a fair bit to that, here (`_attempt_init_of_python_3_9_hash_object`) and in tests - it may be more prudent do that in a separate PR.

Repiteo: I'll do a quick pass to see just how invasive that would be

Repiteo: Looks like all the relevant 3.9-split code was added in #4047, which indeed added a LOT. Unsure if that *needs* a separate PR, but it would be *removing* a ton of code outright instead of just exposing a simple conditional, so it might be warranted? I'm fine crossing that bridge when we come to it

mwichmann:
When this one wakes back up, ActionTests.py will need to resync - shouldn't be too complicated, but it's changed since the PR was created.

SCons/Util/hashes.py
bdbaddog: I'd say make that clean up a separate PR.

Repiteo:
With 4.11's release, I'll consider this ready for review. Updated the relevant documentation to denote 3.9 as the new minimum

bdbaddog:
> With 4.11's release, I'll consider this ready for review. Updated the relevant documentation to denote 3.9 as the new minimum

We'll be pushing out a 4.11.1 release to fix an issued introduced in that release, then this will get merged.

PR #4830, opened 2026-02-17
Split `SCons.Util` even more, discourage implicit imports

Repiteo:
Makes `SCons.Util` more package-y with a couple of changes:
- Disable type hints for `SCons.Util`'s implicit imports. This keeps the functionality working as-is, but now intellisense and IDEs will discourage not using the correct package explicitly.
- Split tests across the different packages, instead of having everything consolidated in `UtilTests.py`.

## Contributor Checklist:


SCons/Util/__init__.py
```diff
@@ -452,8 +457,8 @@ def do_flatten(  # pylint: disable=redefined-outer-name,redefined-builtin
     sequence,
     result,
     isinstance=isinstance,
-    StringTypes=StringTypes,
-    SequenceTypes=SequenceTypes,
+    StringTypes=StringTypes, # type: ignore
+    SequenceTypes=SequenceTypes, # type: ignore
```
Repiteo, line 461: Trying to use `SCons.Util.sctypes.*` here causes circular imports. Could maybe be circumvented with local import syntax, but it's simpler to just use the legacy values and silence the warnings.

mwichmann:
I don't have a problem with this in concept. Away from big screens for a while, so is kind of hard to review.  The new Tests.py files are all moved content, right? (UtilsTests has such a big diff it's collapsing it and not showing, means that'll be hard to tell visually).

Repiteo:
It's all moved content, correct

SCons/Util/__init__.py
```diff
@@ -700,22 +705,22 @@ def WhereIs(file, path=None, pathext=None, reject=None) -> str | None:
                 path = os.environ['PATH']
             except KeyError:
                 return None
-        if is_String(path):
+        if SCons.Util.sctypes.is_String(path):
```
bdbaddog: Why are you making this change? So intellisence will work?

Repiteo: For intellisense, yes. However, that was a bit silly when I could've just suppressed the warnings, so I did that instead

Repiteo:
I'm no longer 100% convinced that this is the best solution, so I'll be marking this as a draft for now. Will move the test changes to another PR

SCons/Util/envs.py
```diff
@@ -74,13 +74,13 @@ def PrependPath(
     orig = oldpath
     is_list = True
     paths = orig
-    if not is_List(orig) and not is_Tuple(orig):
+    if not SCons.Util.sctypes.is_List(orig) and not SCons.Util.sctypes.is_Tuple(orig):
```
bdbaddog, line 77: Can you also remove SCons.Util.'s from this file?

Repiteo:
In the interest of debloating PR count, I'll just close this outright. If I want to revisit the idea, it'll be in the form of a new PR

PR #4833, opened 2026-03-02
Clean up test framework module docstrings

mwichmann:
Also a couple of typing additions and docstring changes in TestSCons. No functional changes.

## Contributor Checklist:

### DIFF
--- a/testing/framework/TestCommon.py
+++ b/testing/framework/TestCommon.py
@@ -1,4 +1,7 @@
+# SPDX-License-Identifier: PSF-2.0
+#
 # Copyright 2000-2010 Steven Knight
+# Copyright The SCons Foundation
 #
 # This module is free software, and you may redistribute it and/or modify
 # it under the same terms as Python itself, so long as this copyright message
@@ -16,6 +19,9 @@
 # SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 #
 # Python License: https://docs.python.org/3/license.html#psf-license
+#
+# This copy is vendored for the SCons project from a revision marked
+# "TestCommon.py 1.3.D001 2010/06/03 12:58:27 knight"
 
 """
 A testing framework for commands and scripts with commonly useful error handling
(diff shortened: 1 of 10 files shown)


testing/framework/TestCommon.py
```diff
@@ -16,6 +19,9 @@
 # SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 #
 # Python License: https://docs.python.org/3/license.html#psf-license
+#
```
bdbaddog: What's this from?

bdbaddog:
Is it worth adding the subclass of subclass of?  If we generate API docs via Sphinx won't it explain that automaticaly?

mwichmann:
> Is it worth adding the subclass of subclass of? If we generate API docs via Sphinx won't it explain that automaticaly?

maybe not worth it... but we don't currently produce API docs of testfw.

testing/framework/TestCommon.py
mwichmann: Not sure the question... that was there ("under the same terms as Python itself") and we haven't gotten it reclicensed to be under the same terms as the rest of SCons.

bdbaddog: this bit

```
# This copy is vendored for the SCons project from a revision marked
# "TestCommon.py 1.3.D001 2010/06/03 12:58:27 knight"```

mwichmann: an attempt to explain that we've continued to change it, while it still has an old version string in it. Our current version isn't "1.3.D001", we couldn't even produce such a version from git...  you can see those lines elsewhere in the two files.

bdbaddog: I think we can just remove those. My guess it was the old "__VERSION"... string which got populated and checked in when it shouldn't have.

mwichmann: happy to remove.  those two layers are made to seem as if they were developed indepenedently and just included in SCons as one of multiple projects that used it. If there ever were others I doubt they still are, so we could recast that whole tone - but I certainy wasn't going to presume that much.

bdbaddog: If I had to guess.. some of this code originated with QMtest which was the original framework used. But I could wrong and this code is the result of removing the need to have QMtest..

PR #4838, opened 2026-03-11
Do not duplicate Tool class creation

mwichmann:
If "default" is part of the tool list (including common case of implicit inclusion when no tools are declared), the `default.py` tool will be run. This calls a function to generate the default tool list, which as a side effect creates a `Tool` instance for each tool examined. The tool then loops through the returned list, creating a fresh `Tool` instance for each tool.  To alleviate this duplication, the `Tool` instances are returned in the default list rather than the tool name string. If an attempt is made to instantiate a Tool with the `name` argument being an existing instance, it is returned as-is, else the normal instantion takes place as before.

In the default tool - use the routine the `SCons.Platform` module provides for getting the default tool list (`DefaultToolList`).  It does the same thing as `default.generate` did directly, but this way "we're using the API" and not duplicating an existing routine.

## Contributor Checklist:

### DIFF
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -55,6 +55,7 @@ import sys
 import os
 import importlib.util
 
+import SCons.Action
 import SCons.Builder
 import SCons.Errors
 import SCons.Node.FS
@@ -288,7 +289,7 @@ class Tool:
                 kw.update(call_kw)
             else:
                 kw = self.init_kw
-        env.AppendUnique(TOOLS=[self.name])
+        env.AppendUnique(TOOLS=[str(self.name)])
         if hasattr(self, 'options'):
             import SCons.Variables
             if 'options' not in env:
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -133,7 +133,7 @@ class Tool:
     without further setup.
 
     .. versionchanged:: NEXT_RELEASE
-       Accept an exsiting instance at creation time and don't duplicate it.
+       Accept an existing instance at creation time and don't duplicate it.
     """
 
     def __new__(cls, name, toolpath=None, **kwargs) -> None:
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -289,7 +289,7 @@ class Tool:
                 kw.update(call_kw)
             else:
                 kw = self.init_kw
-        env.AppendUnique(TOOLS=[str(self.name)])
+        env.AppendUnique(TOOLS=[self.name])
         if hasattr(self, 'options'):
             import SCons.Variables
             if 'options' not in env:
--- a/SCons/Tool/__init__.py
+++ b/SCons/Tool/__init__.py
@@ -22,16 +22,31 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-"""SCons tool selection.
+"""SCons tool subsystem.
 
-Looks for modules that define a callable object that can modify a
-construction environment as appropriate for a given tool (or tool chain).
+Tool specification modules are callable objects that modify construction
+environments to dynamically enable specific types of builds. This module
+provides the support for handling tool modules:
 
-Note that because this subsystem just *selects* a callable that can
-modify a construction environment, it's possible for people to define
-their own "tool specification" in an arbitrary callable function.  No
-one needs to use or tie in to this subsystem in order to roll their own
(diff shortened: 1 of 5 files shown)


SCons/Tool/__init__.py
```diff
@@ -671,18 +711,34 @@ def InstallVersionedLib(self, *args, **kw):
...
-
-    return list(filter(ToolExists, tools))
+        if not SCons.Util.is_String(tool):
+            return tool.exists(env)
+        t = Tool(tool)
+        return t.exists(env)
 
+    results = []
+    for tool in tools:
+        if not SCons.Util.is_String(tool):
```
bdbaddog, line 735: Can this be anything other than a string or a Tool object?

mwichmann:
Well, it's Python... but I don't see how, no.

bdbaddog:
Any work outstanding to take the "Draft" label off this one?

mwichmann:
No. I've made the change and added the changelog snip.

mwichmann:
oops, had to force-push a rebase as the just-added CHANGES and RELEASE things conflicted - master had moved on.

SCons/Tool/__init__.py
```diff
@@ -254,7 +289,7 @@ def __call__(self, env, *args, **kw) -> None:
                 kw.update(call_kw)
             else:
                 kw = self.init_kw
-        env.AppendUnique(TOOLS=[self.name])
+        env.AppendUnique(TOOLS=[str(self.name)])
```
bdbaddog, line 292: Is this needed? isn't .name already a string?

mwichmann: hmm, probably you're right. I'll check eventually.

mwichmann: Reverted.

mwichmann:
Had one of those branch-out-of-sync problems that didn't want to resolve, so made this edit through the github UI.

SCons/Tool/__init__.py
```diff
@@ -306,7 +306,18 @@ def __str__(self) -> str:
...
+        A Tool is equal to another Tool with the same name, and to a
+        string matching its name (so a Tool can be used interchangeably
+        with its name in membership tests and sets). Comparison against
+        any other type is deferred (returns ``NotImplemented``).
+        """
+        if isinstance(other, Tool):
+            return self.name == other.name
+        if isinstance(other, str):
+            return self.name == other
+        return NotImplemented
```
mwichmann, line 320: If you do this, the method annotation is no longer correct.

bdbaddog: I asked the AI about that.. apparently mypy any pyright will handle NotImplemented even if bool is the specified return value.. thoughts?

```
In CPython's typeshed (the canonical type stubs that mypy/pyright use), the object.__eq__ signature is:

  def __eq__(self, value: object, /) -> bool: ...

  It's annotated -> bool — even though object.__eq__ and virtually every real __eq__ can return NotImplemented. Typeshed deliberately does not write -> bool |
  type(NotImplemented).

  Why they do it that way:

  1. NotImplemented is a protocol sentinel, not a normal return value. When a.__eq__(b) returns NotImplemented, the Python interpreter intercepts it and tries the
  reflected b.__eq__(a), ultimately falling back to identity. Code essentially never does result = a.__eq__(b) and then inspects result — you write a == b, which the
  interpreter resolves to a real bool. So from a caller's perspective, the effective type of == is bool, and that's what typeshed annotates.
  2. The honest "full" type is ugly and rarely used. type(NotImplemented) is types.NotImplementedType (exposed as typing-friendly only in 3.10+). Writing -> bool |
  NotImplementedType is technically more precise but typeshed chose bool for ergonomics, and the type checkers special-case NotImplemented in __eq__/__lt__/etc. so
  returning it from a -> bool-annotated dunder is not flagged as an error.

  So "correct per typeshed" means: the established convention — the one mypy and pyright are tuned to expect — is to annotate __eq__ as -> bool and let the type
  checker handle the NotImplemented special-case. My proposed -> bool | type(NotImplemented) was more literally accurate but against that convention, and it's also
  slightly off as a type expression (type(NotImplemented) is a runtime value, not the clean typing form types.NotImplementedType).

  Bottom line: the existing -> bool is already the conventional, checker-approved annotation. There's nothing to fix on __eq__. My earlier framing of -> bool |
  type(NotImplemented) as "more correct" was misleading — I'd recommend leaving it as -> bool.
```

mwichmann: Huh. So "this has happened before" (unsurprisingly) and there's a special case for it. I didn't know that.

bdbaddog: (at least according to claude.. ;) I'll double check independently..  hmm other sources seem to say the same, but the mypy docs don't explicitly say the above.
I think it's probably ok as is.

PR #4840, opened 2026-03-21
Remove qt3 tool

mwichmann:
Qt3 has been deprecated since SCons 4.3. The tool module and the dummy un-versioned module are removed, as well as the tests.  Support code from the test framework and the import test is also removed.

Note there are references to qt in documentation and tests, but used strings in examples (and tests), rather than as a reference to the tool itself, so those were not edited. 

## Contributor Checklist:


mwichmann:
This CI run is seeing the same problem as in #4938 which is presumably a mismatch between the new appveyor setup and the actual provisionong.  Or something.

This is for:  Image: Visual Studio 2017; Environment: WINPYTHON=Python312

* First a version check shows Python 3.12 is not installed
* A request to install it via Chocolatey is made
* That install aborts itself because it finds information that "python312 v3.12.10 already installed"
* The build aborts with a subsequent command, finding:
```
C:\\Python312\\python.exe : The term 'C:\\Python312\\python.exe' is not recognized as the name of a cmdlet, function, 
script file, or operable program.
```

I'm presuming that there's a 3.12, but it's not at that path? (It is indeed not part of Appveyor's own base - the VS 2017 image is on Windows Server 2016, which stops at Python 3.8.

PR #4842, opened 2026-03-25
Assorted typing cleanups

mwichmann:
Fix various niggly typing and checker complaints.

`Environment.py`:

* To avoid modifying a dict-like thing while iterating over it, we did `.copy()`, making an awkward chained construct that checkers don't like. Simpler (and the more common idiom) is to make a list from the keys view, that way you're iterating over that new list and not the dict (and save creating a dict copy).
* Another use of `copy()` was in detecting a `CPPDEFINES` deque, where we want to keep a copy of it and not modify the original - use the deque constructor for this instead of calling copy() method.
* A few complaints about calling `subst` with a bad type were due to us defeating type narrowing - checkers don't know that `is_String`  evaluating true means the object is a string type. Added casts for certain instances of this.
* Align the names of vars in overridden dunder methods in `BuilderDict` with parent `UserDict` for more consistency (one of ours used the same name as a different arg in the parent). Similar for the `JSONEncoder`  `default` method.
* `arg2nodes` `node_factory` argument *could* be passed a `None`, indicate that in the signature.
* Multiple uses of `CacheDir` (it was both an import from `CacheDir` for type checking, and the name of a function in `Environment.py`). This was confusing to at least one type checker, so just use the qualified name.
* The environment `Base` class `_update` and `_update_onlynew` methods were annotated to take an `EnvironmentBase` argument, but in fact they're designed to update the internal dict from a passed dict.  Adjusted the annotation.
* Used the global `Environment` hook directly by name rather than fully qualified, as `SCons.Environment` doesn't have an import of itself.  That's namespace niggling, because it worked as written.
* The `Repository` method (public API level) was written to take kwargs and pass those on to the FS layer's `Repository` - which doesn't take kwargs. `Repository()` was never documented as accepting any keyword arguments (and no tests ever poked at that usage), so just removed to align the two.
* In `VariantDir`, make a small readability change and also avoid reassigning the function parameters: the index-zero access occurs on the call to `self.fs.VariantDir`, not on the end of the call to `arg2nodes`.

`Node/FS.py`: `Glob` and `glob` annotation adjusted to show the exclude argument could also be a single string, as that could be passed down from the public API level, as documented: "The optional exclude argument may be set to a pattern or a list of patterns describing files or directories to filter out of the match list."

`Tool/FortranCommon.py`: fix an error where returns were annotated `Tuple`, which is not imported; changed to `tuple`.

`Util/__init__.py`: use a typevar for `semi_deepcopy`, which is a dispatch function, type checkers don't seem to read through to the return types of the various copy functions it calls.

No test changes - for the small code changes, we want to show behavior *didn't* change.

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -66,6 +66,8 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
       via the SConsignFile(name=None) call.
     - Test framework: tweak module docstrings
     - Test suite: end to end tests don't use assert in result checks
+    - A few more typing cleanups in Environment (and in one case which
+      affected it in the Node package).
 
 
 RELEASE 4.10.1 - Sun, 16 Nov 2025 10:51:57 -0700
--- a/SCons/Environment.py
+++ b/SCons/Environment.py
@@ -85,7 +85,6 @@ from SCons.Util.sctypes import _null
 
 if TYPE_CHECKING:
     from SCons.Action import ActionBase
-    from SCons.CacheDir import CacheDir
     from SCons.Executor import Executor
     from SCons.Node import Node, NodeInfoBase
     from SCons.Node.FS import DirNode, EntryNode, FileNode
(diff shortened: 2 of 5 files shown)


Repiteo (approved):
Everything seems to check out!

Will need `RELEASE.txt` updated, though

mwichmann:
> Everything seems to check out!
> 
> Will need `RELEASE.txt` updated, though

I dodged it as there are already three entries there that collectively say Environment was tweaked for typing and docstrings. *I* figure it's covered... will wait to hear.

PR #4868, opened 2026-06-26
Edits to AGENTS.md file

mwichmann:
These are agent-suggested updates (OpenCode with Qwen 3.6), plus some edits of mine to the result.

In general terms, it suggested removing thing which were instructions to human readers, and some other boilerplate already covered in files it was going to read anyway (README).

I mainly filled out the instructions on doc building, because a question I asked left an agent now knowing how to validate a manpage change.

This has no effect on SCons itself - no docs, code or tests are modified.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,88 +1,99 @@
 # SCons - Software Construction Tool
 
-## Project Overview
-SCons is an open-source software construction tool (build tool) implemented in Python. It is designed to be easier to use and more reliable than the traditional `make` utility. SCons configuration files are Python scripts, allowing users to use the full power of Python to solve build problems.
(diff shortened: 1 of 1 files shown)


.github/workflows/runtest.yml
```diff
@@ -65,7 +65,6 @@ jobs:
             echo "test/ninja/shutdown_scons_daemon.py" >> .excludes
             echo "test/scons-time/run/option/verbose.py" >> .excludes
           elif [[ "${{matrix.os}}" == "windows"* ]]; then
```
bdbaddog, line 67: I'm guessing this was accidently left in this PR? Can you restore to original?

PR #4869, opened 2026-06-27
Subst cleanups

mwichmann:
* Line breaks added for clarity
* Code-style enforced in some places
* In SpecialAttrWrapper.escape, an attribute is read once and stored in a local variable.
* If all previous parts of an if-else chain have done return (or raise), there's no need for the next to be an "else". This causes one fairly large indentation change in StringSubber.expand(), as about 40 lines shift left one indent (checker observation)
* In subst-dict the local variable is renamed from dict to substitutions to avoid "hiding" built-in dict (frequent checker grumble)

No changed behavior, so existing tests and docs are unchanged.

## Contributor Checklist:

### DIFF
--- a/SCons/Subst.py
+++ b/SCons/Subst.py
@@ -54,8 +54,7 @@ def raise_exception(exception, target, s):
     msg = f"{name} `{exception}' trying to evaluate `{s}'"
     if target:
         raise SCons.Errors.BuildError(target[0], msg)
-    else:
-        raise SCons.Errors.UserError(msg)
+    raise SCons.Errors.UserError(msg)
 
 
 class Literal:
@@ -65,6 +64,7 @@ class Literal:
     be interpreted as literal.  When passed to the command interpreter,
     all special characters will be escaped.
     """
+
     def __init__(self, lstr) -> None:
         self.lstr = lstr
 
@@ -144,6 +144,7 @@ class CmdStringHolder(UserString):
     will be escaped, quoted, or neither depending on the value of
     *literal* and the string contents.
     """
+
     def __init__(self, cmd, literal: bool = False) -> None:
         super().__init__(cmd)
         self.literal = literal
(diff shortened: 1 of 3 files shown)


SCons/Subst.py
```diff
@@ -160,23 +161,24 @@ def escape(self, escape_func, quote_func=quote_spaces) -> str:
         After calling this function, the next call to str() will
         return the escaped string.
         """
-        if self.is_literal():
-            return escape_func(self.data)
-        elif ' ' in self.data or '\t' in self.data:
-            return quote_func(self.data)
-        else:
-            return self.data
+        data = self.data
```
bdbaddog, line 164: why data = self.data instead of just using self.data?
same question why self.literal instead of self.is_literal() ?

mwichmann: "data": to avoid as many as three attribute lookups

mwichmann: Oh, missed the second Q. `is_literal` seemed an extra call that wasn't needed:
```py                             
    def is_literal(self) -> bool:
        return self.literal
```

Both would only have a minimal performance improvement, so if you don't want we can just drop the changes.

bdbaddog: There's a a couple .. def is_literal from class Literal() which returns True presently..
can those get called by this logic?

mwichmann: there are things that call `is_literal`  but no `is_literal` method is removed so I don't see any impact. Was that the question?

bdbaddog: not all objects with is_literal() have literal properties, if they are called by this logic, wouldn't it raise an exception?

mwichmann: How's it going to get there in a way to cause a problem?  Somebody *could* inherit from `CmdStringHolder` *and* write their own `__init__` that doesn't initialize `self.literal`, *and* inherit the `escape` method that looks at that. Nothing does that currently, and I'm not quite sure why anybody would want to.

mwichmann: If you don't end up comfortable, just flip it back to `if self.is_literal()`, it's not really worth getting hung up on.

bdbaddog: ok finally had some time to more thoroughly understand and you are correct that it's safe. Sorry for the confusion.

PR #4871, opened 2026-07-01
New test and fixes for AlwaysBuild when using ninja execution engine.

### LINKED ISSUE #4870: AlwaysBuild not honored when using ninja execution engine
**Describe the bug**
The AlwaysBuild command for a target is not honored when using the experimental ninja execution engine.

**Required information**
https://discord.com/channels/571796279483564041/1521525108106330122/1521525112195780709

* Version of SCons
Version 4.10.2, HEAD of master. However I know this goes back to the original ninja release.

* Version of Python
Python 3.14.5, although I know it also fails on 3.6 and 3.10

python.org

* How you installed SCons
This happens both on using “scons-local” as well as from the source tree.

Linux - Ubuntu 20.04, Ubuntu 24.04, Alpine 3.24.1

Here is a simple sconstruct using test2.cpp from test/ninja/ninja-fixture

```
SetOption('experimental','ninja')
DefaultEnvironment(tools=[])
env = Environment()
env.Tool('ninja')

# Compile source to object
obj = env.Object('test2.cpp')
# Force rebuild of object file every time
env.AlwaysBuild(obj)

# Link object to executable
env.Program(target='always_build', source=obj)
``` 

* How you invoke scons (The command line you're using "scons --flags some_arguments”)

When using `python3 <path_to_worktree>/scripts/scons.py` it will build the app and then on subsequent calls will rebuild the .o file per the AlwaysBuild on the target.

When using `python3 <path_to_worktree>/scripts/scons.py --experimental=ninja` it will build the app and then on subsequent calls there is no work to do (which is wrong).

pmova7173630:
Provide a new end-to-end test to verify that AlwaysBuild works correctly with the ninja execution engine. Provide fixes for that support. Updates to the required text files to document the fixes.

## Contributor Checklist:


SCons/Tool/ninja_tool/NinjaState.py
```diff
@@ -783,6 +788,9 @@ def action_to_ninja_build(self, node, action=None):
             if callable(node_callback):
                 node_callback(env, node, build)
 
+        if node.always_build:
+            build["inputs"] = build.get("inputs", []) + ["SconsAlwaysBuildPhony"]
```
mwichmann: Would `'implicit'` be better than `'inputs'` here? (it's a question - I don't actually know)

CHANGES.txt
```diff
@@ -14,6 +14,9 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
 
   From John Doe:
     - Whatever John Doe did.
+  
+  From Paul Movall:
```
bdbaddog: Please add your blurb order by last name among contributors to this release.

SCons/Tool/ninja_tool/NinjaState.py
pmova7173630: According to the ninja documentation, `’implicit’` would be a better choice here. I have modified locally and verified the behavior still is correct, and also verified the appropriate changes to the build.ninja file.

From the ninja documentation:
This is for expressing dependencies that don’t show up on the command line of the command; for example, for a rule that runs a script that reads a hardcoded file, the hardcoded file should be an implicit dependency, as changes to the file should cause the output to rebuild, even though it doesn’t show up in the arguments.

pmova7173630: Fixed in latest commit.

CHANGES.txt
pmova7173630: Fixed in latest commit.

PR #4884, opened 2026-07-26
Eliminate xml table conflict

mwichmann:
This is an AI analysis of errors reported by Apache fop, in this style:
```
[ERROR] FOUserAgent - Invalid property value encountered in column-width="proportional-column-width(1)": org.apache.fop.fo.expr.PropertyException: file:/home/mats/github/scons/build/doc/user/scons-user.fo:12435:16: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width' (See position 12435:95)
```

A conflict arises when building PDFs of docs - `informaltable` elements default to proportional columns, `fop` doesn't support these, so it falls back to fixed widths which are incompatible with some Docbook syntax - this isn't SCons' problem per se, just a hole we fall into because of using `fop`, plus wrapping `lxml` - something that wasn't the default until quite a bit after this code was merged in the first place.

Various approaches were explored - some work in this direction had been partially done, as `$DOCBOOK_XSLTPROCFLAGS` provides a way to pass flags through, but while the docbook tool recognizes that in general, it doesn't actually parse it when the lxml wrapping is used: `parampass` is initialized to an empty dict but never populated with parse results - it now has a comment to that effect in case it turns up needed in future. lxml usage wasn't the default before, it became so after the older approach was retired by upstream. Fixing that the "obvious" way, by adding the flags in each doc SConstruct and parsing the value, opens up some other holes, so instead a fairly simplminded postprocessing step is added to the docbook tool.

Assisted-by: OpenCode 1.18.5 driving Qwen 3.86

## Remove this paragraph

## Contributor Checklist:

### DIFF
--- a/doc/user/SConstruct
+++ b/doc/user/SConstruct
@@ -15,7 +15,7 @@ env = Environment(
     ENV={'PATH': os.environ['PATH']},
     tools=['docbook', 'gs', 'zip'],
     toolpath=['../../SCons/Tool'],
-    # DOCBOOK_XSLTPROCFLAGS="--stringparam  fop.extensions  1",
+    # DOCBOOK_XSLTPROCFLAGS="--stringparam fop.extensions 1",
     DOCBOOK_DEFAULT_XSL_HTML='html.xsl',
     DOCBOOK_DEFAULT_XSL_HTMLCHUNKED='chtml.xsl',
     DOCBOOK_DEFAULT_XSL_PDF='pdf.xsl',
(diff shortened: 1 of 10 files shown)


doc/user/SConstruct
```diff
@@ -42,9 +40,6 @@ has_gs = False
 if env.WhereIs('gs'):
     has_gs = True
 
-#
-# Create the EPUB format
```
bdbaddog, line 46: why remove this comment?

mwichmann: The code seems pretty clear on what it's doing, I didn't think we needed to repeat it (in the pointless cleanup category, I admit).

PR #4887, opened 2026-07-29
Prefer clang over gcc on FreeBSD and OpenBSD

vassilit:
Those platforms ships Clang in their base system.

GCC is somehow [supported](https://www.freebsd.org/status/report-2025-07-2025-09/gcc/), but lag behind.

For the record, to test on FreeBSD, I had to:
- install `py312-build` and `py312-setuptools`,
- install package `gmake` and change all hard-coded occurences of 'make' to 'gmake' (SCons does not support $MAKE convention).
- install `py312-sphinx-book-theme` and `py312-pillow` (for rst2pdf, required even if the build log shows `SKIP_PDF=1`).

## Contributor Checklist:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,11 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
   From John Doe:
     - Whatever John Doe did.
 
+  From Vassili Tchersky:
+    - Add 'bsd' platform definition.
+    - The default tool search order now prefers the LLVM/clang toolchain
+      over GCC on FreeBSD and OpenBSD (PR #4887).
+
   From Cornelii Sandberg:
     - Fix Variables() not reading a saved-variables file (e.g. custom.py)
       from the source directory when building in a separate variant
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -47,6 +47,8 @@ CHANGED/ENHANCED EXISTING FUNCTIONALITY
 - List modifications to existing features, where the previous behavior
   wouldn't actually be considered a bug
 
+- The tool search order now prefers clang over gcc on FreeBSD and OpenBSD.
+
 FIXES
 -----
 
--- a/SCons/Platform/__init__.py
+++ b/SCons/Platform/__init__.py
@@ -84,6 +84,8 @@ def platform_default():
             return 'aix'
         elif 'darwin' in sys.platform:
             return 'darwin'
+        elif sys.platform.startswith(('freebsd', 'openbsd')):
+            return 'bsd'
         else:
             return 'posix'
     elif os.name == 'os2':
--- a/SCons/Platform/bsd.py
+++ b/SCons/Platform/bsd.py
@@ -1,0 +1,35 @@
+# MIT License
+#
+# Copyright The SCons Foundation
+#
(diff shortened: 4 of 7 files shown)


vassilit:
Related to #4886

vassilit:
I think (I have not seen them) that I overlooked side-effects of creating a new platform. There are others `platform == 'posix'` checks in the code.

vassilit:
Not too much.

SCons/Platform/__init__.py
```diff
@@ -84,6 +84,10 @@ def platform_default():
             return 'aix'
         elif 'darwin' in sys.platform:
             return 'darwin'
+        elif sys.platform.startswith('freebsd'):
+            return 'bsd'
```
bdbaddog: make this an or instead of 2 elifs?

CHANGES.txt
```diff
@@ -15,6 +15,10 @@ RELEASE  VERSION/DATE TO BE FILLED IN LATER
   From John Doe:
     - Whatever John Doe did.
 
+  From Vassili Tchersky:
```
bdbaddog, line 18: also added bsd platform definition

SCons/Platform/__init__.py
mwichmann: I'd go for `if 'bsd' in sys.platform` - if it ever turns out they have to return different strings, it can be split later.

vassilit: The thing is Debian GNU/kFreeBSD returns `gnukfreebsd10`, and as a GNU platform, gcc is the default compiler, only the kernel is bsd, not the userland SCons uses.

vassilit: Also NetBSD and DragonFlyBSD have gcc as their base compiler.
MidnightBSD has Clang, but it's quite niche so I'm not sure we should add it.

mwichmann: > Also NetBSD and DragonFlyBSD have gcc as their base compiler. MidnightBSD has Clang, but it's quite niche so I'm not sure we should add it.

Okay, that's reasonable - didn't know these extra details.

vassilit:
One error is:
```
The license for Visual Studio has expired.
Your license for this product is expired and could not be refreshed. Please connect to the internet to continue to use this product.
```

The other one, I'm not sure. I've looked at TestSCons.run, TestCommon.run, TestCmd.run,  etc
But I don't think the modifications I've made have anything to do with the stack size.
Are you sure all the testsuite is idempotent, including the GitHub CI environement ?

mwichmann:
don't worry about these. the visual studio one is an artifact of the ci image at appveyor. since it frozen, they have to make a new one with the silly license refreshed. we heard they had, but not so in practice. the other is one of a small handful of intermittent fails. not your fault.

RELEASE.txt
```diff
@@ -47,6 +47,8 @@ CHANGED/ENHANCED EXISTING FUNCTIONALITY
 - List modifications to existing features, where the previous behavior
   wouldn't actually be considered a bug
 
+- The tool search order now prefers clang over gcc on FreeBSD and OpenBSD.
+
```
bdbaddog, line 51: Please add a blurb about adding bsd platform definition ( around line 122 )

PR #4899, opened 2026-08-31
Ai manpage

bdbaddog:
Added bin/scons-man-to-ai.py which translates the xml manpage into a minimized markdown file suitable for AI to use. (reduces from 230k (scons.1) or 337k (scons-man.html) to 113k tokens with scons-ai.md 

Should this be built and checked into doc/man/scons-ai.md, or only built into build/doc/MD/scons-ai.md?

In the process of this work, found that there were a few variablelist's inside para's where there was nothing else in the para's so simplified those, to also simplify the markdown generation script

(Built with Claude Opus 5)

## Contributor Checklist:


mwichmann:
Guess it's okay. Since the result isn't human-friendly, nor human-intended, it's hard to say much about it.

How would this be provided?  you planning to put it in doc/man even though it's "generated"?

bdbaddog:
> Guess it's okay. Since the result isn't human-friendly, nor human-intended, it's hard to say much about it.
> 
> How would this be provided? you planning to put it in doc/man even though it's "generated"?

not sure. the .gen files are under doc/generated..  if we put it there and then referenced it in our agents.md maybe?

mwichmann:
> > Guess it's okay. Since the result isn't human-friendly, nor human-intended, it's hard to say much about it.
> > How would this be provided? you planning to put it in doc/man even though it's "generated"?
> 
> not sure. the .gen files are under doc/generated.. if we put it there and then referenced it in our agents.md maybe?

I'd say that makes sense.

mwichmann:
Have let this soak for a while, and I'm still not sure it's needed... the current agents are pretty good at picking out relevant snippets from a big file without trying to consume the whole thing. UpToYou, of course.

bdbaddog:
From what I understand they're *very* good at markdown, and it's simpler to parse, but might be even better if we broke it up into sections as separate files and just had a top level TOC with links. The key is keeping the context window size down..  I don't see any way it would hurt..

I should add a pointer to this from the AGENTS.md as well and then merge it.

PR #4904, opened 2026-09-14
ParseDepends: keep Windows drive-letter paths in depfiles

r3wretrhy:
`Environment.ParseDepends` split each make-style depfile line on the first colon. A Windows line such as `C:\build\foo.o: C:\src\foo.c` was therefore treated as target `C` and the rest as dependencies.

Skip colons that are drive letters (`C:\` / `C:/`) and keep the existing `target.split()` / `depends.split()` behaviour for Unix-style files.

The added unit test fails on the unpatched tree and passes with the change.

Tested: `python runtest.py SCons/EnvironmentTests.py`


mwichmann:
This seems a valid concern.

bdbaddog:
Looks good to me. I just pushed a minor change for the new function where I changed the comment on the functions functionality to be a docstring.
Merging.

